PR #4, opened 2016-07-15
Cleaning up builds and jobs

stjohnjohnson:
- More clear on what environment is
- Removing triggers and triggered_by as we need a better way to represent workflow
- Removing duplicate fields
- Removing executor as we don't expose that field from the executor
- Adding job description as an optional field


model/job.js
```diff
@@ -14,7 +11,12 @@ const MODEL = {
     name: Joi
         .string()
         .description('Name of the Job')
-        .example('component'),
+        .example('main'),
+
+    description: Joi
+        .string()
```
nkatzman: Should we set a max size?

d2lam:
Can we bump the major version?

model/task.js
```diff
@@ -18,9 +18,10 @@ const MODEL = {
         .description('Container this Task is running in')
         .example('node:4'),
 
-    environment: Joi
-        .object()
-        .description('Key=>Value environment variables for the task itself'),
+    environmentId: Joi
```
d2lam: Just confirming: does your other PR handle the order of env? For example: we expect 
`{a:1; b:2}` and `{b:2; a:1}` to have the same environmentId.

stjohnjohnson: It should, it sorts the keys and hashes them.

model/job.js
stjohnjohnson: Good catch.

stjohnjohnson:
Bumped major, limited description, and added limit for job name.

nkatzman:
👍  seems good

model/job.js
```diff
@@ -12,9 +9,15 @@ const MODEL = {
         .example('50dc14f719cdc2c9cb1fb0e49dd2acc4cf6189a0'),
 
     name: Joi
-        .string()
+        .string().regex(/^[A-Za-z0-9_-]+$/).min(3)
```
d2lam: Any reason why we want to make jobname >= 3 characters?

stjohnjohnson: I was thinking more around how domain names are minimum lengthed.  Anything under 3 doesn't provide enough context.

Actually, domain names minimum is 2 characters: https://www.quora.com/What-is-the-minimum-number-of-characters-for-a-domain-name

Do you think this is unnecessary?

d2lam: If I'm a lazy user and want to create some quick job to test out screwdriver and use "dp" instead of deployment, this might annoy me. I think it's kinda unnecessary but I'm not super against it.

tkyi:
👍  lgtm

stjohnjohnson:
Moving to DNM pending further discussion about tasks

test/data/build.yaml
```diff
@@ -1,3 +1,16 @@
 # Build Example
+id: 8843d7f92416211de9ebb963ff4ce28125932878
 jobId: 50dc14f719cdc2c9cb1fb0e49dd2acc4cf6189a0
-runNumber: 15
+parentBuildId: d8fd39d0bbdd2dcf322d8b11390a4c5825b11495
+number: 15
```
tkyi: so would the build number be `d8fd39d0bbdd2dcf322d8b11390a4c5825b11495.15` in this case?

tkyi:
👍 LGTM

d2lam:
👍

PR #5, opened 2016-07-21
Define tableName and indexes

nkatzman:
This PR defines attributes in the data-schema that can be used for automated the creation of tables in a datastore

The models that we need to create tables for now all contain tableName and indexes to create

### DIFF
--- a/model/pipeline.js
+++ b/model/pipeline.js
@@ -79,5 +79,21 @@ module.exports = {
      * @property keys
      * @type {Array}
      */
-    keys: ['scmUrl']
+    keys: ['scmUrl'],
+
+    /**
+     * Tablename to be used in the datastore
+     *
+     * @property tableName
+     * @type {String}
+     */
+    tableName: 'pipelines',
+
+    /**
+     * List of indexes to create in the datastore
+     *
+     * @property indexes
+     * @type {Array}
+     */
+    indexes: ['scmUrl']
 };
--- a/model/build.js
+++ b/model/build.js
@@ -119,5 +119,21 @@ module.exports = {
      * @property keys
      * @type {Array}
      */
-    keys: ['jobId', 'number']
+    keys: ['jobId', 'number'],
+
+    /**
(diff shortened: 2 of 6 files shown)


d2lam:
LGTM 👍

FenrirUnbound:
This is great for defining which index to use for each table. Do we have any future plans for other criteria, like Local Secondary Index in DynamoDB?

model/pipeline.js
```diff
@@ -79,5 +79,21 @@ module.exports = {
...
-    keys: ['scmUrl']
+    keys: ['scmUrl'],
+
+    /**
+     * Tablename to be used in the datastore
+     *
+     * @property tableName
+     * @type {String}
+     */
+    tableName: 'pipelines',
```
tkyi: Are we going with `Pipelines` or `pipelines`? (https://github.com/screwdriver-cd/dynamic-dynamodb/pull/1/files#diff-64fbbb5a69337f996786fef9c19e1271R23)

model/pipeline.js
```diff
@@ -79,5 +79,21 @@ module.exports = {
...
+
+    /**
+     * List of indexes to create in the datastore
+     *
+     * @property indexes
+     * @type {Array}
+     */
+    indexes: ['scmUrl']
```
tkyi: Is this the `IndexName` or `AttributeName`?

model/pipeline.js
nkatzman: My vote is for all lowercase plurals i.e. 'pipelines'

model/pipeline.js
nkatzman: this is meant to be the attributeName to create the global index on

nkatzman:
@FenrirUnbound are there situations where we want a secondary index? Are those things that are useful outside of DynamoDB?

stjohnjohnson:
👍  
mother__have_you_gone_mad

jer:
👍

PR #7, opened 2016-07-22
Increasing Data Schema's Scope to include sd.yaml and validator IO

stjohnjohnson:
- Major version bump for backwards incompatible changes
- index now exposes access to config, api, and model
- Added examples for all config and api validation

PR #8, opened 2016-07-22
Add validation for datastore

d2lam:
To validate inputs for the datastore

### DIFF
--- a/model/datastore.js
+++ b/model/datastore.js
@@ -1,0 +1,67 @@
+'use strict';
+const Joi = require('joi');
+const table = Joi.string().required();
+const id = Joi.string().hex().required();
+const idReferenceObject = Joi.object().keys({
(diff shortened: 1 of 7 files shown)


stjohnjohnson:
DNM until https://github.com/screwdriver-cd/data-schema/pull/7 plz

model/datastore.js
```diff
@@ -0,0 +1,67 @@
...
+const dataReferenceObject = Joi.object().keys({
+    table,
+    params: Joi.object().keys({
+        id,
+        data: Joi.object().required()
+    }).required()
+});
+const paginateReferenceObject = Joi.object().keys({
+    table,
+    params: Joi.object(),
```
nkatzman: do we do anything with params right now? Should we just add this in when we start doing params?

PR #10, opened 2016-07-22
Add validation for executor

d2lam:
Keeping STREAM and STOP separate in case we need to add more things to STOP later. 

I think STOP only needs buildId but IDK since it's not implemented yet... 

Resolves https://github.com/screwdriver-cd/executor-base/issues/9


nkatzman:
👍

stjohnjohnson:
👍

PR #12, opened 2016-07-23
Labels are throwing off the validator, removing them

stjohnjohnson:
For complex structures like the config, label isn't helping. It's saying Job is broken instead of "main"

With labels:
```js
const Joi = require('joi');
const stepSchema = Joi.string().alphanum().label('Step');
const stepsSchema = Joi.object().keys({
    init: stepSchema,
    test: stepSchema,
    publish: stepSchema
}).label('List of Steps');

const example = {
    init: 'foo',
    test: 'BAD-THINGS_HERE!!!',
    publish: 'bar'
};
Joi.validate(example, stepsSchema).error.toString();
// 'ValidationError: child "Step" fails because ["Step" must only contain alpha-numeric characters]'
```

Without labels:
```js
const Joi = require('joi');
const stepSchema = Joi.string().alphanum();
const stepsSchema = Joi.object().keys({
    init: stepSchema,
    test: stepSchema,
    publish: stepSchema
});

const example = {
    init: 'foo',
    test: 'BAD-THINGS_HERE!!!',
    publish: 'bar'
};
Joi.validate(example, stepsSchema).error.toString();
// 'ValidationError: child "test" fails because ["test" must only contain alpha-numeric characters]'
```

### DIFF
--- a/config/job.js
+++ b/config/job.js
@@ -11,7 +11,6 @@ const SCHEMA_MATRIX = Joi.object()
     // All others are marked as invalid
     .unknown(false)
     // Add documentation
-    .label('Job Matrix')
     .options({
         language: {
             object: {
@@ -28,7 +27,6 @@ const SCHEMA_ENVIRONMENT = Joi.object()
     // All others are marked as invalid
     .unknown(false)
     // Add documentation
-    .label('Environment')
     .options({
         language: {
             object: {
@@ -44,7 +42,6 @@ const SCHEMA_STEPS = Joi.object()
     // All others are marked as invalid
     .unknown(false)
     // Add documentation
-    .label('Steps')
     .options({
         language: {
             object: {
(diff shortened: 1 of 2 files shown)


FenrirUnbound:
So if I'm understanding this correctly, it's using the label "Job". And because it's using the label, it the error too vague.

This may lead us into a longer conversation, and we can table it if it comes to it. Since you're proposing to remove the labels, how would the validator determine which job is the problematic one?

stjohnjohnson:
@FenrirUnbound Great point!  In my haste last night I completely gave way too little context here.  I've updated the description with a chunk of code to reproduce the problem I encountered.

d2lam:
👍 Makes sense

FenrirUnbound:
Thanks for the clarification!

👍  Looks great to me

jkepuv9

PR #13, opened 2016-07-25
Add token to user-schema

d2lam:


d2lam:
Might not need this actually

PR #14, opened 2016-07-27
Removing Platforms until we come up with the Template design

stjohnjohnson:
Since we're certainly not going to be adding the concept of "platforms" and instead going to replace it with "templates", let's just remove all the Platform code and simplify the interfaces.


d2lam:
👍

d2lam:
👍

d2lam:
Need to change dynamo-datastore as well 
https://github.com/screwdriver-cd/datastore-dynamodb/blob/master/index.js#L11

And delete the table.

nkatzman:
bump version please

nkatzman:
👍

PR #15, opened 2016-07-27
Add token to user base

d2lam:
Login is not working because validation failed. 
Token is currently not part of the user base so we need to add it.


Filbird:
= 👍 )

nkatzman:
👍

PR #17, opened 2016-07-27
Adding JobName to start (this is really only temporary)

stjohnjohnson:
This is only temporary until we migrate to just passing `buildId`.


d2lam:
👍

PR #20, opened 2016-08-04
Build Webhook Api Validation

nkatzman:
This PR is adding in api validation for build webhooks. 

When a build has finished we want to update status and meta


petey:
👍 

cho_cho_

jer:
👍

PR #21, opened 2016-08-05
Add a property containing a list of all the keys in the model

petey:
Since our models should be a representation of a database row, it sure would be nice to be able to get all the field names.

### DIFF
--- a/models/build.js
+++ b/models/build.js
@@ -122,6 +122,13 @@ module.exports = {
     keys: ['jobId', 'number'],
 
     /**
+     * List of all fields in the model
+     * @property allKeys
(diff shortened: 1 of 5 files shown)


nkatzman:
Where do we plan on using this field?

petey:
@nkatzman The base model should be able to take in the return data from datastore.* and convert that to the keys in the model. In order to do that, it needs to know what those keys are.

stjohnjohnson:
👍 
mrw_i_visit_a_new_sub_and_sort_by_top_all_time

FenrirUnbound:
👍 

kdkhgx6

PR #23, opened 2016-08-08
Switching to a new screwdriver.yaml and validator format

stjohnjohnson:
This has a new format for the validator (to be json valid):

```json
{"name": "install", "command": "npm install"}
```

It also changes the screwdriver.yaml with a new step format:
```yaml
steps:
  - npm install
  - install: npm install
```

### DIFF
--- a/api/validator.js
+++ b/api/validator.js
@@ -5,22 +5,35 @@ const Regex = require('../config/regex');
 const Job = require('../config/job');
 const Workflow = require('../config/workflow');
 
+const SCHEMA_JOB_COMMAND = Joi.object()
+    .keys({
+        name: Joi.string(),
+        command: Joi.string()
+    })
+    .unknown(false)
+    .label('Named command to execute');
+
+const SCHEMA_JOB_COMMANDS = Joi.array()
+    .items(SCHEMA_JOB_COMMAND)
+    .min(1)
+    .label('List of named commands to execute');
+
 const SCHEMA_JOB_PERMUTATION = Joi.object()
     .keys({
         image: Job.image,
-        steps: Job.steps,
+        commands: SCHEMA_JOB_COMMANDS,
         environment: Job.environment
     }).label('Job permutation');
 
-const SCHEMA_JOB_PERMUATIONS = Joi.array(SCHEMA_JOB_PERMUTATION)
+const SCHEMA_JOB_PERMUTATIONS = Joi.array().items(SCHEMA_JOB_PERMUTATION)
     .label('List of job permutations');
 
 const SCHEMA_JOBS = Joi.object()
     .keys({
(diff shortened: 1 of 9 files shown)


api/validator.js
```diff
@@ -5,14 +5,27 @@ const Regex = require('../config/regex');
...
 const SCHEMA_JOB_PERMUTATION = Joi.object()
     .keys({
         image: Job.image,
-        steps: Job.steps,
+        steps: SCHEMA_JOB_COMMANDS,
         environment: Job.environment
     }).label('Job permutation');
 
-const SCHEMA_JOB_PERMUATIONS = Joi.array(SCHEMA_JOB_PERMUTATION)
+const SCHEMA_JOB_PERMUATIONS = Joi.array().items(SCHEMA_JOB_PERMUTATION)
```
d2lam: Not sure how this get passed before. Typo here
`const SCHEMA_JOB_PERMUTATIONS`

jer:
👍

test/data/validator.output.yaml
```diff
@@ -2,30 +2,40 @@ jobs:
   main:
   - image: node:4
     steps:
-      install: npm install
-      test: npm test
-      build: npm run build
+      - name: install
```
d2lam: Any reasons why we want to switch to this new format?

jer: Javascript, YAML, JSON, and Go all don't guarantee key ordering for hashes. We absolutely have to have an array in the JSON output, and we decided that the Yaml should be written in an unambiguous way. This also leaves the door open for adding more metadata to steps over time.

nkatzman:
👍

test/data/validator.output.yaml
d2lam: Ok. Makes sense

jer:
YAML keys are unordered in maps: http://yaml.org/spec/1.0/#id2558578
Recommendation for ordering keys in YAML: http://yaml.org/spec/1.0/#id2491203
Golang currently randomizes map key ordering: https://blog.golang.org/go-maps-in-action#TOC_7.

config/job.js
```diff
@@ -49,6 +52,8 @@ const SCHEMA_STEPS = Joi.object()
             }
         }
     });
+const SCHEMA_STEP = Joi.alternatives().try(SCHEMA_STEP_STRING, SCHEMA_STEP_OBJECT);
```
d2lam: So step can still be just a string like `npm publish` 
What will the validator output be for the `name`?

jer: That's coming in the next commit. It's going to be: step-<index>

d2lam: Cool!

d2lam:
👍

PR #26, opened 2016-08-16
Add container to job schema

d2lam:
Job schema should have container so when we create builds, we can look up this information


FenrirUnbound:
👍 

Looks good to me

3lxt4bi - imgur

nkatzman:
👍

petey:
👍

PR #29, opened 2016-08-17
date to isoDate

nicolaifsf:
The way we currently format our timestamps is the date-time format, however according to our own swagger spec, we state that this is simply the date format. this PR suggests a fix that would bring us to accordance with the swagger spec with the way we send timestamps.

### DIFF
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -20,7 +20,8 @@ const MODEL = {
         .example('git@github.com:screwdriver-cd/optional-config.git#master'),
 
     createTime: Joi
-        .date()
+				.string()
+        .isoDate()
         .description('When this pipeline was created'),
 
     admins: Joi
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -20,7 +20,7 @@ const MODEL = {
         .example('git@github.com:screwdriver-cd/optional-config.git#master'),
 
     createTime: Joi
-				.string()
+	.string()
         .isoDate()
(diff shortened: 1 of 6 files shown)


d2lam:
👍

models/pipeline.js
```diff
@@ -20,7 +20,8 @@ const MODEL = {
         .example('git@github.com:screwdriver-cd/optional-config.git#master'),
 
     createTime: Joi
-        .date()
+				.string()
```
jer: Something weird happened to the formatting here

nicolaifsf: hum. this is not the first time this has happened to me. 
I'll fix it.

jer:
So is this a bug in hapi-swagger that is preventing it from defining the right json from .date()? It seems nicer to keep it a date object instead of holding on to a string if we can get away with it, but if this is the only way to make the swagger work, I'm 👍 until the underlying issue is resolved.

Also please squash the commits and force-push back if you could

nicolaifsf:
That's a really good question. Specifically wether or not it actually is a bug or not.
So, hapi-swagger in their examples actually uses this string().isoDate() method in order to follow the ISO8601 standard for date-time formatting (how we currently present the dates) https://github.com/hapijs/joi/blob/master/test/string.js#L2567-L2609
But after doing some digging right now to answer your question, I found that in joi .date().iso() appears to have the same effect in specifying the standard that we are using.

PR #30, opened 2016-08-17
Optional ref parameter

nkatzman:
When reading a file from source-control it should be possible to pass a reference, i.e. commit, branch, or tag, that will be used for reading the file


d2lam:
👍

shruthi-venkateswaran:
👍

PR #31, opened 2016-08-17
date to isoDate

nicolaifsf:
changed from .date() to .string().isoDate() so codegen will work and we ensure compliance with swagger spec for dateTime formatting. http://swagger.io/specification/ and definition of date-time standard. http://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14

### DIFF
--- a/models/build.js
+++ b/models/build.js
@@ -40,15 +40,18 @@ const MODEL = {
         .example('ccc49349d3cffbd12ea9e3d41521480b4aa5de5f'),
 
     createTime: Joi
-        .date()
+        .string()
+        .isoDate()
         .description('When this build was created'),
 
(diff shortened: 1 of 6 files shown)


tkyi:
👍  LGTM

jer:
👍  Do it!

nkatzman:
Should this bump the major version?

FenrirUnbound:
Is it backwards compatible?

jer:
This should be backward-compatible. It *should* be restricting the values to the format that we are already using, but just guaranteeing iso8601 format now. I vote we merge it as-is and revert if we hit a snag.

nkatzman:
👍  lets merge then

PR #33, opened 2016-08-23
Adding Log Lines, Step Details, and Permutations

stjohnjohnson:
- Starting a new Log Lines model
- Adding step details to Build
- Adding Permutations to Job

### DIFF
--- a/models/job.js
+++ b/models/job.js
@@ -1,6 +1,7 @@
 'use strict';
 const Joi = require('joi');
 const mutate = require('../lib/mutate');
+const validator = require('../api/validator');
 
 const MODEL = {
     id: Joi
@@ -19,6 +20,8 @@ const MODEL = {
         .items(Joi.string())
         .description('List of container images')
         .example(['node:4', 'node:6']),
+
+    permutations: validator.jobPermutations,
 
     description: Joi
         .string().max(100)
@@ -58,7 +61,7 @@ module.exports = {
     get: Joi.object(mutate(MODEL, [
         'id', 'pipelineId', 'name', 'state'
     ], [
-        'description', 'containers'
+        'description', 'containers', 'permutations'
     ])).label('Get Job'),
 
     /**
@@ -68,8 +71,7 @@ module.exports = {
      * @type {Joi}
      */
     update: Joi.object(mutate(MODEL, [], [
-        'state',
-        'containers'
+        'state'
     ])).label('Update Job'),
 
     /**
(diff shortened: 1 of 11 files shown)


models/log.js
```diff
@@ -0,0 +1,84 @@
...
+    base: Joi.object(MODEL).label('Log'),
+
+    /**
+     * Properties for Log that will come back during a GET request
+     *
+     * @property get
+     * @type {Joi}
+     */
+    get: Joi.object(mutate(MODEL, [
+        'id', 'buildId', 'timestamp', 'message', 'number', 'stepName', 'time'
```
minz1027: Why 'timestamp' is here? Only 'time' is defined in the MODEL.

minz1027: Why 'timestamp' is here? Only 'time' is defined in the MODEL.

models/job.js
```diff
@@ -68,8 +71,7 @@ module.exports = {
      * @type {Joi}
      */
     update: Joi.object(mutate(MODEL, [], [
-        'state',
-        'containers'
+        'state'
```
d2lam: we don't allow update of containers anymore?

stjohnjohnson: We don't allow update of builds other than status.

nkatzman: This is for jobs?

Should containers only ever be updated through the screwdriver.yaml?

stjohnjohnson: Correct.

models/log.js
```diff
@@ -0,0 +1,84 @@
...
+     */
+    tableName: 'logs',
+
+    /**
+     * List of indexes to create in the datastore
+     *
+     * @property indexes
+     * @type {Array}
+     */
+    indexes: ['buildId', 'stepName']
```
minz1027: In the doc, the Log API will `get step#1 logs from lineNumber` to get logs from db. If we only index `buildId` and `stepName`, does it mean we need to scan all lines for the specified step to find the lines we want?

stjohnjohnson: Good point, I'm really not sure what happens with such a large number of indexes.  @nkatzman

minz1027:
LGTM 👍

test/data/build.yaml
```diff
@@ -14,3 +14,20 @@ parameters:
 meta:
     coverage: 99.54
 status: SUCCESS
```
petey: This really doesn't look like a successful job based on the status codes in steps.

stjohnjohnson: Heh

petey:
👍  
stjohn aliens

d2lam:
👍

PR #34, opened 2016-08-24
Export the right fields for Steps

stjohnjohnson:
This is so we can use the steps fields for `/builds/:id/steps/:name`

Feature: https://github.com/screwdriver-cd/screwdriver/issues/135


petey:
👍

api/loglines.js
```diff
@@ -0,0 +1,42 @@
...
+}).label('URL Parameters');
+
+const SCHEMA_QUERY = Joi.object().keys({
+    from: Joi.number().integer()
+        .min(0).default(0)
+        .description('Starting Line Number')
+}).label('Query Parameters');
+
+const SCHEMA_LOGLINE = Joi.object().keys({
+    n: Joi
```
d2lam: So the emitter writes to the log service with the log line as `{t,m,s}` https://github.com/screwdriver-cd/launcher/blob/e5c4cb1048152ca20773cca62617fe36cc9f9194/screwdriver/emitter.go#L25-L29

Are we going to modify these values in the log service or something?

stjohnjohnson: Yup

jer:
👍

PR #35, opened 2016-08-26
I'm positive that 0 is not a positive line number. :(

stjohnjohnson:
Since our line numbers start at 0, we need to remove this.


Filbird:
Looks positive 👍

tkyi:
👍 makes no sense

PR #37, opened 2016-08-29
Removing Log Schema (we're not using it)

stjohnjohnson:
Since we decided to bypass storing logs in the Datastore, we don't need this schema anymore.


Filbird:
👍 
giphy-1

d2lam:
👍  LGTM

PR #38, opened 2016-08-29
Adding more commit information to Build

stjohnjohnson:
Including:
 - Who started the build (link to user info)
 - Who authored the commit (link to user info)
 - Commit message
 - Link to Commit

### DIFF
--- a/models/build.js
+++ b/models/build.js
@@ -47,13 +47,43 @@ const MODEL = {
     cause: Joi
         .string()
         .description('Reason why this build started')
-        .example('Commit ccc493 was pushed to master'),
+        .example('pull request opened'),
 
     sha: Joi
         .string().hex()
         .length(40)
         .description('SHA this project was built on')
         .example('ccc49349d3cffbd12ea9e3d41521480b4aa5de5f'),
+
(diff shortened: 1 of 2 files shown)


models/build.js
```diff
@@ -47,14 +47,44 @@ const MODEL = {
...
 
     sha: Joi
         .string().hex()
         .length(40)
         .description('SHA this project was built on')
         .example('ccc49349d3cffbd12ea9e3d41521480b4aa5de5f'),
 
+    startedBy: Joi
+        .string()
+        .description('Identier of who started this build')
```
Filbird: Identifier* :^)

Filbird:
👍

d2lam:
Code looks fine. 👍  
But is that too much information? I imagine `startedBy` and `authoredBy` would be enough. Do we really need `startedByUri` and `authoredByUri`? Isn't it just going to be `http://github.com/${username}`

stjohnjohnson:
@d2lam the point of this is so that we don't hard code github.com.  What if you're running your own GHE instance or we want to support GitLab.

tkyi:
👍

PR #40, opened 2016-09-01
Allow meta to be passed during update

stjohnjohnson:
This allows us to get rid of /webhooks/build and replace it with PUT /builds/:id


petey:
👍  metatron

PR #44, opened 2016-09-03
Removing isPR as it's a function in Models

stjohnjohnson:
Currently breaking Screwdriver release as they conflict.


jer:
👍 makes sense

d2lam:
Ah this one

PR #45, opened 2016-09-08
Removing containers (not used anymore) and adding state as an index

stjohnjohnson:
We need the index so we can just get the list of active jobs.


tkyi:
👍  seems reasonable

PR #46, opened 2016-09-08
Secret value should be optional for GET

d2lam:


stjohnjohnson:
Good catch, we only want to return the value if a build is asking for it. 👍

PR #48, opened 2016-09-13
Removing Webhooks and Adding SCM

stjohnjohnson:
Build Webhook was removed in https://github.com/screwdriver-cd/screwdriver/pull/153

Adding SCM for https://github.com/screwdriver-cd/scm-base/pull/5


core/scm.js
```diff
@@ -0,0 +1,75 @@
...
+        .example('https://github.com/stjohnjohnson'),
+
+    name: Joi.string()
+        .required()
+        .label('Display Name')
+        .example('stjohnjohnson'),
+
+    avatar: Joi.string()
+        .uri()
+        .required()
```
d2lam: This seems specific for github. Do we know other source control would have an "avatar" for users?

core/scm.js
```diff
@@ -0,0 +1,75 @@
...
+        .label('Link to Repository')
+        .example('https://github.com/screwdriver-cd/screwdriver/tree/master')
+}).label('SCM Repository');
+
+const SCHEMA_COMMIT = Joi.object().keys({
+    sha: Joi.string()
+        .hex()
+        .length(40)
+        .required()
+        .label('SHA1 identifier for the commit')
```
d2lam: `SHA` instead of `SHA1`

stjohnjohnson: But it is a SHA1, anything else won't be 40 characters.

core/scm.js
stjohnjohnson: Pretty much all of them: 
 - https://docs.gitlab.com/ee/customization/libravatar.html
 - https://blog.bitbucket.org/2013/07/16/customize-your-avatar-in-bitbucket-no-gravatar-needed/

d2lam:
Yeah I'm fine with this 👍

FenrirUnbound:
👍 

Looks good to me

this is fine l - imgur

PR #50, opened 2016-09-14
Define sorting key for build model

FenrirUnbound:
Define a primary sorting key for the `build` model.

When you perform a query for a set of build models, they are in a non-specific order. This defines a column to sort the data by.


stjohnjohnson:
gifs

PR #51, opened 2016-09-15
Allow sort in the scan config

d2lam:
Add `sort` to the allowed keys. This prepares for the datastore change so that we can sort

Solves https://github.com/screwdriver-cd/screwdriver/issues/187

### DIFF
--- a/api/pagination.js
+++ b/api/pagination.js
@@ -11,7 +11,13 @@ const MODEL = {
         .number().integer().min(1)
         .max(50)
         .default(50)
-        .description('Count to paginate')
+        .description('Count to paginate'),
+
+    sort: Joi
+        .string().lowercase()
+        .valid(['ascending', 'descending'])
+        .default('descending')
+        .description('Sorting option')
 };
 
 /**
--- a/plugins/datastore.js
+++ b/plugins/datastore.js
@@ -15,13 +15,14 @@ const SCHEMA_DATA = Joi.object().keys({
         data: Joi.object().required()
     }).required()
 });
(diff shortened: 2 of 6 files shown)


plugins/datastore.js
```diff
@@ -15,13 +15,14 @@ const SCHEMA_DATA = Joi.object().keys({
...
     table,
     params: Joi.object(),
     paginate: Joi.object().keys({
-        count: Joi.number().required(),
-        page: Joi.number().required()
-    }).required()
+        count: Joi.number().integer().positive().required(),
+        page: Joi.number().integer().positive().required()
+    }).required(),
+    sort: Joi.string().lowercase().valid(['ascending', 'descending']).optional()
```
stjohnjohnson: Can we change this from optional to required and default it to something?

FenrirUnbound: Defaulted it to `descending`. So it'll order by newest build (e.g., most recent) to oldest build

d2lam: Need to check what happens if we sort on a table that doesn't have sort key

d2lam: Tested locally. It's fine to sort on table that does not have a sort key. It will just not sort.

api/pagination.js
```diff
@@ -11,7 +11,13 @@ const MODEL = {
         .number().integer().min(1)
         .max(50)
         .default(50)
-        .description('Count to paginate')
+        .description('Count to paginate'),
+
+    sort: Joi
```
FenrirUnbound: 👍

PR #52, opened 2016-09-15
Define multiple rangeKeys

FenrirUnbound:
This PR addresses the case where we need to have multiple range keys -- or fewer range keys than indices. We cannot decorate the `indexes` property, since the `rangeKey` property is specific to only one datastore (dynamodb).

Blocked by #51 
Solves https://github.com/screwdriver-cd/screwdriver/issues/187

### DIFF
--- a/models/build.js
+++ b/models/build.js
@@ -208,11 +208,13 @@ module.exports = {
 
     /**
      * Primary column to sort queries by.
-     * This defines queries to optionally sort a query result set by build number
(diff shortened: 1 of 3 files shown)


d2lam:
👍  Don't think this needs to be blocked by #51. I will just resolve the merge conflict if this is merged.

tkyi:
👍 seems reasonable to me

PR #53, opened 2016-09-16
[1] Add "archive" attribute to job model

FenrirUnbound:
This PR adds an archive attribute to the job model. The purpose of this additional field is to flag a specific job as active or archived. Related to screwdriver-cd/screwdriver#188

Jobs that are associated with open PRs workflow jobs are considered to be active. 

Jobs that are associated with closed/merged PRs or are jobs used outside of the workflow are considered to be archived.

### DIFF
--- a/models/job.js
+++ b/models/job.js
@@ -34,7 +34,13 @@ const MODEL = {
         ])
         .description('Current state of the Job')
         .example('ENABLED')
-        .default('ENABLED')
+        .default('ENABLED'),
+
+    archived: Joi
+        .boolean()
+        .description('Flag if the job is archived')
+        .example(true)
+        .default(false)
 };
 
 module.exports = {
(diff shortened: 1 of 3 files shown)


models/job.js
```diff
@@ -34,7 +34,13 @@ const MODEL = {
         ])
         .description('Current state of the Job')
         .example('ENABLED')
-        .default('ENABLED')
+        .default('ENABLED'),
+
+    archive: Joi
```
stjohnjohnson: It might sound better if it's `archived`

models/job.js
```diff
@@ -34,7 +34,13 @@ const MODEL = {
...
+        .boolean()
+        .description('Flag if the job is inactive')
```
d2lam: Maybe use `archived` instead of `inactive` here

d2lam:
The rest LGTM 👍

PR #55, opened 2016-09-19
[1] Adding optional jobName to updateStatus

stjohnjohnson:
This let's us put multiple messages on a given commit.


FenrirUnbound:
We continually add things that are "optional". Do we intend on testing a case where no optional parameters are used?

PR #56, opened 2016-09-20
Allow to store scmRepo object to pipeline

catto:
This PR allows to store scmRepo object to a pipeline. 
(screwdriver-cd/screwdriver#194 and screwdriver-cd/screwdriver#160)

### DIFF
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -77,10 +77,10 @@ module.exports = {
      * @type {Joi}
      */
     create: Joi.object(mutate(MODEL, [
-        'scmUrl',
-        'scmRepo'
+        'scmUrl'
     ], [
-        'configUrl'
+        'configUrl',
+        'scmRepo'
     ])).label('Create Pipeline'),
 
     /**
@@ -89,7 +89,7 @@ module.exports = {
      * @property keys
      * @type {Array}
      */
-    keys: ['scmRepo.id'],
+    keys: ['scmUrl'],
 
     /**
      * List of all fields in the model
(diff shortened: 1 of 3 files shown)


catto:
bumped the version to 13

models/pipeline.js
```diff
@@ -74,7 +77,8 @@ module.exports = {
      * @type {Joi}
      */
     create: Joi.object(mutate(MODEL, [
-        'scmUrl'
+        'scmUrl',
+        'scmRepo'
```
stjohnjohnson: It's not required for create.

models/pipeline.js
```diff
@@ -85,7 +89,7 @@ module.exports = {
      * @property keys
      * @type {Array}
      */
-    keys: ['scmUrl'],
+    keys: ['scmRepo.id'],
```
stjohnjohnson: Don't change the key yet, that way it won't be backwards incompatible.  We can slowly migrate towards using it.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "screwdriver-data-schema",
-  "version": "12.1.0",
+  "version": "13.0.0",
```
stjohnjohnson: If we remove the key change, this is not-backwards incompatible and we can set this back to `12.1.0`

models/pipeline.js
catto: I moved it to optionalFields

models/pipeline.js
catto: Reverted this and version

PR #58, opened 2016-09-21
Fixing the pipeline model for scmRepo and removing configUrl

stjohnjohnson:
We have 0 code in our system for `configUrl` right now, so I'm removing it until we want to bring it back.

Additionally, we need to return `scmRepo` as part of the GET and we dont' want users to pass it to us during create, it's a auto-generated value.


stjohnjohnson:
Soon.

PR #61, opened 2016-09-23
Change the key for pipeline to scmRepo.id

catto:
This PR changes the key for generating pipeline ID. 
screwdriver-cd/screwdriver#160


d2lam:
This LGTM 👍  I guess we would need to regenerate the tables?
Spomb7hNzvofC

PR #62, opened 2016-10-03
Add update secret schema

minz1027:
- Add `update` schema to secret to allow update on `value` or/and `allowInPR`
- Remove `remove` schema from secret since we are not using it

First PR in  [Issue #232](https://github.com/screwdriver-cd/screwdriver/issues/232)


d2lam:
👍 LGTM

PR #63, opened 2016-10-03
Revert catto's change for now

minz1027:
- Revert @catto 's [change](https://github.com/screwdriver-cd/data-schema/commit/1d72e57df84311358452f34a83bc043db4981057) for now
- We are not ready for it yet

### DIFF
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -79,7 +79,7 @@ module.exports = {
      * @property keys
      * @type {Array}
      */
-    keys: ['scmRepo.id'],
+    keys: ['scmUrl'],
 
     /**
      * List of all fields in the model


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "screwdriver-data-schema",
-  "version": "13.0.0",
+  "version": "12.3.0",
```
FenrirUnbound: Well this is going to be awkward. That means that we have to keep track of the fact that version 13 exists, while the `master` branch shows version 12 is the latest.

minz1027: Keep version 13.0.0 but revert catto's change. Is that alright?

FenrirUnbound: catto's update was valid. why undo it?

FenrirUnbound: Why can't we patch 12.x with what we need?

stjohnjohnson: I think because it's still WIP, we're in a inconsistent state.  I'd revert the change, keep at 13, and open a PR to revert the revert when ready.

PR #67, opened 2016-10-11
Replace scmURL with scmURi and add schema for scm decorator

minz1027:
- Replace `scmUrl` with `scmUri` (e.g. github.com:123456:master) as key
- Add `checkoutUrl` which can be either `ssh` or `https`
- Add schema to validate the input and output for scm plugin methods


core/scm.js
```diff
@@ -68,8 +57,54 @@ const SCHEMA_COMMIT = Joi.object().keys({
...
+    branch: Joi.string()
+        .required()
+        .label('Branch of the repository'),
+
+    prRef: Joi.string()
+        .optional()
+        .label('PR reference of the repository'),
+
+    sha: Joi.string().hex()
+        .length(40)
```
d2lam: bitbucket does not always give back the full sha

core/scm.js
```diff
@@ -68,8 +57,54 @@ const SCHEMA_COMMIT = Joi.object().keys({
...
+        .required()
+        .label('Commit SHA')
+        .example('ccc49349d3cffbd12ea9e3d41521480b4aa5de5f'),
+
+    username: Joi.string()
+        .required()
+        .label('Display Name')
```
d2lam: This should be username, not Display Name. They are different.
d2lam vs Dao Lam

core/scm.js
```diff
@@ -26,15 +27,15 @@ const SCHEMA_USER = Joi.object().keys({
 }).label('SCM User');
 
 const SCHEMA_REPO = Joi.object().keys({
-    id: Joi.string()
+    title: Joi.string()
```
d2lam: I feel like the name `title` and `subtitle` is a little confusing. Any reason why we want to name them like that?

core/scm.js
minz1027: removed the length() now

core/scm.js
minz1027: Fixed

models/pipeline.js
```diff
@@ -11,10 +11,16 @@ const MODEL = {
...
-        .example('git@github.com:screwdriver-cd/data-model.git#master'),
+    checkoutUrl: Joi
+        .string().regex(Regex.CHECKOUT_URL)
+        .description('Checkout url for the application')
+        .example('git@github.com:screwdriver-cd/data-schema.git#master')
+        .example('https://github.com/screwdriver-cd/data-schema.git#master'),
+
+    scmUri: Joi
+        .string().regex(Regex.SCM_URI)
+        .description('Unique idetifier for the application')
```
tkyi: identifier*

PR #72, opened 2016-10-14
Change the scmUri regex to match any string

minz1027:
- Change the scmUri regex to match any string as ID except for `:`


d2lam:
👍

PR #76, opened 2016-10-18
Allow display name to be null

minz1027:
- Display name can be null for github users
image


minz1027:
Close this and set name to username if null in `scm-github`.
https://github.com/screwdriver-cd/scm-github/pull/19

PR #78, opened 2016-10-21
add getFileRef schema for scm.getFile() to take in prRef

minz1027:
-  `scm.getFile(`) will either use `scmUri` or `prRef` to get file
- Add an extra schema for scm.getFile() when it takes in `prRef` instead of `scmUri`

### DIFF
--- a/test/plugins/scm.test.js
+++ b/test/plugins/scm.test.js
@@ -45,6 +45,16 @@ describe('scm test', () => {
         });
     });
 
+    describe('getFileRef', () => {
+        it('validates', () => {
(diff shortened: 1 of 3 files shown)


test/plugins/scm.test.js
```diff
@@ -45,6 +45,16 @@ describe('scm test', () => {
         });
     });
 
+    describe('getFileRef', () => {
+        it('validates', () => {
+            assert.isNull(validate('scm.getFile.yaml', scm.getFile).error);
```
tkyi: scm.getFileRef.yaml**

PR #81, opened 2016-10-24
Add event uuid to scm parseHook schema

minz1027:
- Add event uuid to parseHook schema
- It will be needed in API when we switch to parseHook for getting all event data
- https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/webhooks/github.js#L339

### DIFF
--- a/core/scm.js
+++ b/core/scm.js
@@ -68,6 +68,10 @@ const SCHEMA_HOOK = Joi.object().keys({
         .required()
         .label('Type of the event'),
 
+    hookId: Joi.string()
+        .required()
+        .label('Uuid of the event'),
+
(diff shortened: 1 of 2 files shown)


core/scm.js
```diff
@@ -68,6 +68,10 @@ const SCHEMA_HOOK = Joi.object().keys({
         .required()
         .label('Type of the event'),
 
+    eventId: Joi.string()
```
stjohnjohnson: Just be careful about the name here.  Is this the same as our event IDs?

minz1027: Nah, it is the uuid for a webhook event. changed it to hookId now

PR #82, opened 2016-10-25
[event-1] Define schema for event model

FenrirUnbound:
Include the schema for the `event` model. 

Relates to https://github.com/screwdriver-cd/screwdriver/issues/221

### DIFF
--- a/models/event.js
+++ b/models/event.js
@@ -1,0 +1,100 @@
+'use strict';
+
+const Joi = require('joi');
+const mutate = require('../lib/mutate');
+const Pipeline = require('./pipeline');
+const Scm = require('../core/scm');
+const Workflow = require('../config/workflow');
+
+const MODEL = {
+    id: Joi
+        .string().hex().length(40)
+        .description('Identifier of this build')
+        .example('4b8d9b530d2e5e297b4f470d5b0a6e1310d29c5e'),
+    causeMessage: Joi
+        .string().max(512).truncate()
+        .description('Message that describes why the event was created')
+        .example('Merge pull request #26 from screwdriver-cd/data-schema'),
+    commit: Scm.commit
+        .description('Commit related to the event'),
+    createTime: Joi
+        .string()
+        .isoDate()
+        .description('When this event was created')
+        .example('2038-01-19T03:14:08.131Z'),
+    creator: Scm.user
+        .description('Creator of the event'),
+    pipelineId: Joi.reach(Pipeline.base, 'id')
+        .description('The pipeline that the event belongs to')
+        .example('2d991790bab1ac8576097ca87f170df73410b55c'),
+    sha: Joi
+        .string().hex()
+        .description('SHA this project was built on')
+        .example('ccc49349d3cffbd12ea9e3d41521480b4aa5de5f'),
+    status: Joi
+        .string().valid([
+            'SUCCESS',
+            'FAILURE',
+            'ABORTED',
+            'RUNNING'
+        ])
+        .description('Current status of the event')
+        .example('SUCCESS')
+        .default('RUNNING'),
+    workflow: Workflow.workflow
+        .description('Workflow of the associated pipeline')
+        .example(['main', 'publish', 'deploy'])
+};
+
+module.exports = {
+    /**
+     * List of all fields in the model
+     * @property allKeys
+     * @type {Array}
+     */
+    allKeys: Object.keys(MODEL),
+
+    /**
+     * All the available properties of Event
+     *
+     * @property base
+     * @type {Joi}
+     */
+    base: Joi.object(MODEL).label('Event'),
+
+    /**
+     * Properties for Event that will come back during a GET request
+     *
+     * @property get
+     * @type {Joi}
+     */
+    get: Joi.object(mutate(MODEL, [
+        'id', 'commit', 'createTime', 'creator', 'pipelineId', 'sha', 'status', 'workflow'
+    ], [
+        'causeMessage'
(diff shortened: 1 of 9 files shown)


models/event.js
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * @property create
+     * @type {Joi}
+     */
+    create: Joi.object(mutate(MODEL, [
+        'createTime',
+        'creator',
+        'pipelineId',
+        'sha'
+    ])).label('Create Event'),
```
d2lam: I feel like we won't need this and `update`. We aren't exposing these endpoints to users.

test/data/event.get.full.yaml
```diff
@@ -0,0 +1,24 @@
...
+    message: Fixing a bug with signing
+    author:
+        url: https://github.com/stjohnjohnson
+        name: St. John Johnson
+        username: stjohnjohnson
+        avatar: https://avatars.githubusercontent.com/u/622065?v=3
+    url: https://github.com/scredriver-cd/screwdriver/commit/46f1a0bd5592a2f9244ca321b129902a06b53e03
+createTime: "2038-01-19T03:14:08.131Z"
+creator:
+    name: St. John Johnson
```
d2lam: I know this is test but could we make this a different user to make it more obvious that creator != committer.

models/event.js
FenrirUnbound: 👍

test/data/event.get.full.yaml
FenrirUnbound: 👍

test/data/event.get.yaml
```diff
@@ -0,0 +1,23 @@
+# Event Get Example
+id: bbf22a3808c19dc50777258a253805b14fb3ad8b
+causeMessage: "Manually started by d2lam"
```
d2lam: This probably should be `stjohnjohnson` to make sense since he's the creator.

models/event.js
```diff
@@ -0,0 +1,97 @@
...
+const Workflow = require('../config/workflow');
+
+const buildId = Joi.reach(Build.base, 'id');
+
+const MODEL = {
+    id: Joi
+        .string().hex().length(40)
+        .description('Identifier of this build')
+        .example('4b8d9b530d2e5e297b4f470d5b0a6e1310d29c5e'),
+    builds: Joi
```
d2lam: @stjohnjohnson Yeah we were debating about this. What is your thought? 
Having eventId inside builds is more consistent with our current models. Like we don't have a list of builds inside a job. 
Having builds inside event helps the UI. UI can just hits `/pipelineId/events` to get a list of events, and for each event it can get the list of builds to display.

stjohnjohnson: I assume we're going to move away from `/jobs/:jobId/builds` and into `/events/:eventId/builds`  If that's the case, wouldn't the UI be making less calls?

 - `/pipelines/:pipelineId/events` - list of events
 - `/events/:eventId/builds` - list of builds for that event

models/event.js
```diff
@@ -0,0 +1,97 @@
...
+const Build = require('./build');
+const Joi = require('joi');
+const mutate = require('../lib/mutate');
+const Pipeline = require('./pipeline');
+const Scm = require('../core/scm');
+const Workflow = require('../config/workflow');
+
+const buildId = Joi.reach(Build.base, 'id');
+
+const MODEL = {
```
stjohnjohnson: Do we want to have overall status?  Did the event complete successfully?  Or do we just want to infer that from all builds?

models/event.js
FenrirUnbound: > Like we don't have a list of builds inside a job. 

The number of builds in an event is static when it's created, whereas the number of builds inside a job can vary. 

That being said, this is more of a premature optimization based on what the UI would find convenient to consume. We should drop the `builds` field here and add it in the future as a backwards-compatible change. 

Regardless of what we do with the `builds` field, the build model will still need to reference the event it's associated with.

stjohnjohnson: @FenrirUnbound Eh, if you think this is a better approach, I'm willing to concede here. I was just concerned about an inconsistency in our APIs as well as a race between two builds starting and updating the same row.

FenrirUnbound: > I worry about a possible race condition

The way we were thinking about it was that you can determine which builds are attached to this event when you access it. But, like I said, this is an optimization. You'll still need to query the related builds for the full data set, and we should treat the build table as the source of truth if it comes down to a race condition. 

One of many philosophies with NoSQL usage is that you promote some small amount of redundancy across tables. For instance, the Github API carries some redundancy with its calls by providing some portion of User information in their commit data. 

I'm not saying we should follow suit; it's just difficult to break out of that mindset when your initial datastore is DynamoDB. We have to support SQL-based implementations too, so we can't carry that type of philosophy if we need to support normalized SQL tables.

FenrirUnbound: I would much prefer starting off with less and adding more as time goes on. I'll work on removing this in the next push and if we really need this later down the road, we can add it. Taking it away will be a lot harder

models/event.js
FenrirUnbound: Initially, I wanted to defer with saving the overall status in the event model, just because I'm not sure we need it quite yet. It would be convenient. 

Some of the CD solutions we looked at earlier this year gave us some options on "what is deployment success" (e.g., If half of the machines are deployed to, consider it a success). Maintaining the status in the `event` model would enable that possibility. 

I'll add it in the next update as well.

d2lam:
Merge conflicts

PR #83, opened 2016-10-25
[1] Add ping event type to hook schema, make action optional

tkyi:
Adding the `ping` event to the hook schema since the scm plugin will handle a `ping` event in its `parseHook` method; making `action` field optional since `ping` events do not have an action


core/scm.js
```diff
@@ -75,7 +75,7 @@ const SCHEMA_HOOK = Joi.object().keys({
     action: Joi.alternatives()
         .when('type', { is: 'pr', then: Joi.valid(['opened', 'closed', 'synchronized']) })
         .when('type', { is: 'repo', then: Joi.valid('push') })
-        .required()
+        .optional()
```
petey: Maybe make it conditionally optional?

```
.when('type', { is: 'ping', then: Joi.optional(), otherwise: Joi.required() })
```

FenrirUnbound:
Test?

tkyi:
I could add different mock data but it would just be running the same test?

core/scm.js
```diff
@@ -64,18 +64,18 @@ const SCHEMA_COMMIT = Joi.object().keys({
...
     hookId: Joi.string()
         .required()
         .label('Uuid of the event'),
 
-    action: Joi.alternatives()
+    action: Joi.string()
         .when('type', { is: 'pr', then: Joi.valid(['opened', 'closed', 'synchronized']) })
         .when('type', { is: 'repo', then: Joi.valid('push') })
-        .required()
+        .when('type', { is: 'ping', then: Joi.optional(), otherwise: Joi.required() })
```
minz1027: Cool! 👍

petey:
@FenrirUnbound  brings up a good point. Since behavior is different for different hook calls, we should probably have a test for each hook payload.

https://github.com/screwdriver-cd/data-schema/blob/master/test/core/scm.test.js#L39-L41

tkyi:
@FenrirUnbound bumped

PR #87, opened 2016-11-01
Remove event status

d2lam:
Let's remove event status for now to simplify the logic of the API. 

We can infer the status of an event from the builds' status.


FenrirUnbound:
No version bump?

d2lam:
@FenrirUnbound  I don't think it's necessary. This is more like a bug fix since models was assuming `status` isn't there. This won't break anything.

PR #91, opened 2016-11-08
Add commands and environment to steps in builds

tkyi:
Adding commands and environment information to builds so that the launcher can use this information when running commands


minz1027:
bump minor version?

PR #93, opened 2016-11-10
Add schema to validate input for `getCheckoutCommand` in scm

minz1027:
- We will generate checkout command in SCM plugin with `getCheckoutCommand` method
- This schema is used to validate the input for `getCheckoutCommand` method
Relates to https://github.com/screwdriver-cd/screwdriver/issues/312

### DIFF
--- a/plugins/scm.js
+++ b/plugins/scm.js
@@ -10,6 +10,15 @@ const jobName = Joi.reach(models.job.base, 'name').optional();
 const username = Joi.reach(models.user.base, 'username').required();
 const checkoutUrl = Joi.reach(models.pipeline.create, 'checkoutUrl').required();
 
+const GET_CHECKOUT_COMMAND = Joi.object().keys({
+    branch: Joi.string().required(),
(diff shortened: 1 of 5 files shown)


plugins/scm.js
```diff
@@ -10,6 +10,15 @@ const jobName = Joi.reach(models.job.base, 'name').optional();
 const username = Joi.reach(models.user.base, 'username').required();
 const checkoutUrl = Joi.reach(models.pipeline.create, 'checkoutUrl').required();
 
+const GET_CHECKOUT_COMMAND = Joi.object().keys({
+    branch: Joi.string().required(),
+    host: Joi.string().required(),
+    org: Joi.string().required(),
+    repo: Joi.string().required(),
+    sha: Joi.string().required(),
```
d2lam: Why is sha required?

minz1027: Because we will not checkout with `prRef` directly. We will fetch the pr and merge the pr's `sha` to the latest version of master branch.

PR #94, opened 2016-11-11
[1] Add checkout command output validation for scm plugins

tkyi:
We will require scm-plugins to output `sd-checkout-code` as the name for a checkout command

Related to feedback in https://github.com/screwdriver-cd/models/pull/132

### DIFF
--- a/core/scm.js
+++ b/core/scm.js
@@ -45,6 +45,17 @@ const SCHEMA_REPO = Joi.object().keys({
         .example('https://github.com/screwdriver-cd/screwdriver/tree/master')
 }).label('SCM Repository');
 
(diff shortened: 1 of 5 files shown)


core/scm.js
```diff
@@ -45,6 +45,17 @@ const SCHEMA_REPO = Joi.object().keys({
         .example('https://github.com/screwdriver-cd/screwdriver/tree/master')
 }).label('SCM Repository');
 
+const SCHEMA_COMMAND = Joi.object().keys({
+    name: Joi.string().regex(/^sd-checkout-code$/)
```
d2lam: Hmmm why not just use `Joi.equal('sd-checkout-code')`?

PR #96, opened 2016-12-07
feat: addWebhook method input

FenrirUnbound:
Joi validation for `scm-base.addWebhook`, which will help enforce the input for the method.

Based on the implementations for [`scm-github`](https://github.com/screwdriver-cd/scm-github/pull/45) and [`scm-bitbucket`](https://github.com/screwdriver-cd/scm-bitbucket/pull/29), we determined that the most common interface between them contains:

* `scmUri` - the SCM URI of the repository
* `token` - the auth token to interact with the SCM service
* `url` - the webhook to notify when certain events occur

Related to screwdriver-cd/screwdriver#379

### DIFF
--- a/test/data/scm.addWebhook.yaml
+++ b/test/data/scm.addWebhook.yaml
@@ -1,0 +1,4 @@
+---
+scmUri: github.com:126987453:master
+token: thisisashinytoken
+url: https://screwdriver.cd/v4/some-endpoint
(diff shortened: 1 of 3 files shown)


test/data/scm.addWebhook.yaml
```diff
@@ -0,0 +1,4 @@
+---
+scmUri: github.com:126987453:master
+token: thisisashinytoken
```
tkyi: so shinyyy

PR #102, opened 2017-01-01
fix: Add length for mysql index

catto:
I found the same issue as screwdriver-cd/screwdriver#325 on the latest version of Screwdriver.
I added length limitations for each key to fix it.

### DIFF
--- a/models/event.js
+++ b/models/event.js
@@ -19,6 +19,7 @@ const MODEL = {
     createTime: Joi
         .string()
         .isoDate()
+        .max(32)
         .description('When this event was created')
         .example('2038-01-19T03:14:08.131Z'),
     creator: Scm.user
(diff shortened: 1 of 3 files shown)


models/event.js
```diff
@@ -19,6 +19,7 @@ const MODEL = {
     createTime: Joi
         .string()
         .isoDate()
+        .max(32)
```
catto: Should `createTime` be date, not string?

stjohnjohnson: Good question.  I think we were storing it as a String and letting the users parse it.  But we may want to switch to Date.

catto: I got it 🐈

tkyi:
We're going to wait to merge this PR. Might have to go through and rerun a bunch of other pipelines to pick up the new `data-schema` version with your change after it's merged.

PR #104, opened 2017-01-06
fix: Add field to give a reason why a job is disabled

tkyi:
It would be useful to the user to be able to see why a job is disabled
Adding this field would be a first step towards that feature


d2lam:
Probably not going to implement this feature now as the priority is low. I will close this and we can revisit later.

PR #106, opened 2017-02-23
feat: Add notifications field to pipeline

r3rastogi:
We want to allow users to be notified by email when their Screwdriver
job changes status. Thus, add an optional email notification field to
the pipeline model.

### DIFF
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -35,11 +35,16 @@ const MODEL = {
 
     admins: Joi
         .object()
-        .description('Admins of this Pipeline')
+        .description('Admins of this pipeline')
         .example({ myself: true }),
 
     workflow: Workflow.workflow
-        .description('Current workflow of the pipeline')
+        .description('Current workflow of the pipeline'),
+
+    notificationEmail: Joi
+        .string().email()
+        .description('Email to send notifications to')
+        .example('screwdriver-cd@github.com')
 };
 
 module.exports = {
@@ -60,7 +65,7 @@ module.exports = {
     get: Joi.object(mutate(MODEL, [
         'id', 'scmUri', 'createTime', 'admins'
     ], [
-        'workflow', 'scmRepo'
+        'workflow', 'scmRepo', 'notificationEmail'
     ])).label('Get Pipeline'),
 
     /**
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -24,8 +24,10 @@
     "Dao Lam <daolam112@gmail.com>",
     "Darren Matsumoto <aeneascorrupt@gmail.com>",
     "Jeremiah Wuenschel <jeremiah.wuenschel@gmail.com>",
+    "Jerry Zhang <thejerryzhang@gmail.com>",
```
tkyi: Can you add Min while you're at it? 😄

models/pipeline.js
```diff
@@ -35,11 +35,16 @@ const MODEL = {
...
         .object()
-        .description('Admins of this Pipeline')
+        .description('Admins of this pipeline')
         .example({ myself: true }),
 
     workflow: Workflow.workflow
-        .description('Current workflow of the pipeline')
+        .description('Current workflow of the pipeline'),
+
+    notificationEmail: Joi
```
stjohnjohnson: This is not extendable. 👎

stjohnjohnson:
Also, isn't this supposed to be configurable per job?

models/pipeline.js
r3rastogi: Do you have any suggestions on better ways of storing the email?

r3rastogi:
@stjohnjohnson Do you have any suggestions on better ways of storing the email?

package.json
r3rastogi: Added

minz1027: Thanks @tkyi :)

stjohnjohnson:
@r3rastogi I wouldn't store any of this in the model.  We already store a flattened job config in the job table, you can just refer to that in the model.

We should look at adding the configuration setting, something like:
```yaml
jobs:
  main:
    settings:
      email: foo@example.com
```

d2lam:
@stjohnjohnson  You mean to add another one called `settings` under job (right now it doesn't have it), or you're ok with using: 
```
jobs: 
  main:
     environment: 
        SD_EMAIL: dao@yahoo.com
```

d2lam:
Discussed with stjohn offline. Putting it in `settings` so we don't overload `environment`.

Updated issue here: 
https://github.com/screwdriver-cd/screwdriver/issues/465

r3rastogi:
Will address this feature as described in screwdriver-cd/screwdriver#465

PR #107, opened 2017-02-23
feat: Add settings to job config

jerryzhang222:
Added to config for each job, according to https://github.com/screwdriver-cd/screwdriver/issues/465

The `jobs` config will now allow:

```
jobs: 
   component:
      settings:
             email: email@yahoo.com
```


config/job.js
```diff
@@ -60,13 +60,26 @@ const SCHEMA_STEP_OBJECT = Joi.object()
 const SCHEMA_STEP = Joi.alternatives().try(SCHEMA_STEP_STRING, SCHEMA_STEP_OBJECT);
 const SCHEMA_STEPS = Joi.array().items(SCHEMA_STEP).min(1);
 const SCHEMA_IMAGE = Joi.string();
+const SCHEMA_EMAIL = Joi.string().email();
+const SCHEMA_EMAILS = Joi.array()
```
stjohnjohnson: Why are we requiring an array?

config/job.js
```diff
@@ -60,13 +60,26 @@ const SCHEMA_STEP_OBJECT = Joi.object()
...
+    .items(SCHEMA_EMAIL)
+    .min(0);
+const SCHEMA_NOTIFICATIONS = Joi.object()
```
stjohnjohnson: Why are we using notifications?  I thought we were building a flat model?

config/job.js
```diff
@@ -60,13 +60,26 @@ const SCHEMA_STEP_OBJECT = Joi.object()
...
+    });
 const SCHEMA_JOB = Joi.object()
     .keys({
         steps: SCHEMA_STEPS,
         environment: SCHEMA_ENVIRONMENT,
         matrix: SCHEMA_MATRIX,
         image: SCHEMA_IMAGE,
-        secrets: SCHEMA_SECRETS
+        secrets: SCHEMA_SECRETS,
+        settings: SCHEMA_SETTINGS
```
stjohnjohnson: The more I think about this, the more we need to be careful about this name.  Maybe something like `features`.

config/job.js
d2lam: I think `email` should accept both `array` of emails or  `email`. 
Not 2 different fields `email` or `emails` though.

config/job.js
d2lam: `settings` makes more sense to me actually.
What do other people think?

config/job.js
minz1027: https://github.com/screwdriver-cd/screwdriver/issues/465
If we want to add more features/settings later, the flat model will look confusing or we need to make a long variable name, e.g. notification_email, notification_slack...
```
settings:
     notifications:
     other-feature:
```
wouldn't this look more organized?

config/job.js
jerryzhang222: SCHEMA_EMAILS is an array of Joi's of format SCHEMA_EMAIL (similar to the pattern used for SCHEMA_STEPS -> SCHEMA_STEP and SCHEMA_SECRETS -> SCHEMA_SECRET)

config/job.js
stjohnjohnson: Or just `email`.  Why do we need it to be called `notification_email`.

petey: Why does it need to be `notifications_*`? Why not just `email`, `hipchat`, `slack`, `featureA`?

config/job.js
stjohnjohnson: But what if you want to customize it, like:
```yaml
shared:
  settings:
    email:
      address: foo@example.com
      notify_developers: true
```

How are we going to support the evolving use-cases of new features/settings/plugins thingy.

config/job.js
jerryzhang222: I think it would be more intuitive for a new user to know what their email, hipchat, slack would be used for (notifications). Is there a drawback to having more layers of organization?

config/job.js
jerryzhang222:
```
const SCHEMA_EMAIL = Joi.object()
    .keys({
       address: foo@example.com
       notify_developers: true
    })
```

And `SCHEMA_EMAILS` stays the same?

Would that work?

d2lam: For simplification, Let's make `email` an object. If they want one email, still 
```yaml
email:
     addresses: [foo@example.com]
```

if they want more settings:
```yaml
email:
   addresses: [foo@example.com, bar@example.com]
   notify_developers: true
```

d2lam:
Seems like convention is if it's an object then singular (like `environment`), if it's an array then plural (`secrets`). We should follow that.

petey:
A notification mechanism should be singular, e.g. `email`. Otherwise you will get into some weird territory: `hipchats`, `slacks`. I for one do not want to configure a trouser-based notification system.

d2lam:
@jerryzhang222  Also it's using your yahoo email to commit.

PR #109, opened 2017-02-28
feat(470): Add template validation

tkyi:
We will need to add schemas for what is expected/valid in a ```template.yaml```

Expected/required fields in a `template.yaml`:
- name (eg: chef/cookbook, chef/pipeline-upload, chef/pipeline-assembly, etc.)
- description
- version
- maintainer/point of contact
- config
     - steps

Related to https://github.com/screwdriver-cd/screwdriver/issues/470

TODO (in another PR): add `lifespan` schema

### DIFF
--- a/config/template.js
+++ b/config/template.js
@@ -1,0 +1,46 @@
+'use strict';
+
+const Joi = require('joi');
+const Job = require('./job');
+const Regex = require('./regex');
+
+const TEMPLATE_NAME = Joi
+            .string()
+            .regex(Regex.TEMPLATE_NAME)
+            .description('Name of the Template')
+            .example('node/npm-install');
+
+const TEMPLATE_VERSION = Joi
+            .string()
+            .regex(Regex.VERSION)
+            .description('Version of the Template')
+            .example('1.2.3');
+
+const TEMPLATE_DESCRIPTION = Joi
+            .string()
+            .description('Description of the Template')
+            .example('Installs npm modules');
+
+const TEMPLATE_MAINTAINER = Joi
+            .string()
+            .email()
+            .description('Maintainer of the Template')
+            .example('foo@bar.com');
+
+const SCHEMA_TEMPLATE = Joi.object()
+    .keys({
+        name: TEMPLATE_NAME,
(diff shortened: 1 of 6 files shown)


test/data/config.template.yaml
```diff
@@ -0,0 +1,7 @@
+name: node/install-test
+version: 1.2.0
+maintainer: foo@bar.com
+description: Template to install and test npm packages
+steps:
```
d2lam: Based on stjohn's suggestion, we should have this at `config` level. Can you take a look at the updated issue? Also there are other fields in here.

tkyi: added a `config`
not sure what sort of validation we're expecting for a `lifespan`

config/template.js
```diff
@@ -0,0 +1,55 @@
...
+const TEMPLATE_MAINTAINER = Joi
+            .string()
+            .email()
+            .description('Maintainer of the Template')
+            .example('foo@bar.com');
+
+const TEMPLATE_CONFIG = Joi
+            .object()
+            .keys({
+                steps: Job.steps
```
d2lam: What about other fields?
In fact, since we decided it's going to be similar to `job`, we can reuse this: https://github.com/screwdriver-cd/data-schema/blob/master/config/job.js#L87

tkyi: Is everything in the Job config allowed?

config/template.js
```diff
@@ -0,0 +1,47 @@
...
+const SCHEMA_TEMPLATE = Joi.object()
+    .keys({
+        name: TEMPLATE_NAME,
+        version: TEMPLATE_VERSION,
+        description: TEMPLATE_DESCRIPTION,
+        maintainer: TEMPLATE_MAINTAINER,
+        config: Job.job
+    })
+    .requiredKeys('name', 'version', 'description', 'maintainer', 'config.steps')
+    .default({});
```
minz1027: if we always require all the keys why we need default {} ?

minz1027: just this question, otherwise LGTM 👍

tkyi: good point! 👍 
removed

PR #110, opened 2017-03-02
feat(470): add template schema to models

minz1027:
- Add template schema to models
- Highlight:
   -  `scmUri` is the template pipeline's scmUri
   - `name` and `version` define a unique record
   -   A record (e.g. test-template@1.3) can have multiple `labels`
   -  only `labels` can be updated for a record

Related to https://github.com/screwdriver-cd/screwdriver/issues/470

### DIFF
--- a/models/template.js
+++ b/models/template.js
@@ -1,0 +1,106 @@
+'use strict';
+
+const Joi = require('joi');
+const mutate = require('../lib/mutate');
+const Template = require('../config/template');
+const scmUri = Joi.reach(require('./pipeline').base, 'scmUri');
+
+const MODEL = {
+    id: Joi
+        .number().integer().positive()
+        .description('Identifier of this template')
+        .example(123345),
+
+    labels: Joi
+        .array()
+        .items(Joi.string())
+        .description('Labels for template')
+        .example(['stable', 'latest', 'beta']),
+
+    config: Template.config,
+    name: Template.name,
+    version: Template.version,
+    description: Template.description,
+    maintainer: Template.maintainer,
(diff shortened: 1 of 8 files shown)


models/template.js
```diff
@@ -0,0 +1,111 @@
...
+        .description('Identifier of this template')
+        .example(123345),
+
+    labels: Joi
+        .array()
+        .items(Joi.string())
+        .description('Labels for template')
+        .example(['stable', 'latest', 'beta']),
+
+    config: Joi.reach(Template.template, 'config'),
```
d2lam: Maybe we can just expose those fields in `../config/template`, so you won't need to do `Joi.reach` here multiple times

tkyi: So line 5 could be `const Template = require('../config/template').template;`
So this line would be `config: Template.config,`

models/template.js
```diff
@@ -0,0 +1,106 @@
...
+     */
+    allKeys: Object.keys(MODEL),
+
+    /**
+     * List of indexes to create in the datastore
+     *
+     * @property indexes
+     * @type {Array}
+     */
+    indexes: ['name'],
```
FenrirUnbound: Question: with RDS, are indices defined by the `data-schema`? Or is the schema defined separately?

The reason why I ask is because the interface for declaring a template is `template_name@version`. Given this format, I expect `version` to be queried a lot as well as the `name`.

FenrirUnbound: I guess another way to phrase my question is "What does the rangeKeys value map to in RDS"?

PR #112, opened 2017-03-08
fix(470): require template validator inputs/outputs object structures

FenrirUnbound:
## Context

By default, keys in a `Joi.object()` are optional. This means that if a key is omitted or missing, it will not be validated against the prescribed rules. 

For instance, if the input was given an empty object, it will accept it as-is. We want the input and output to have a reasonably consistent structure.

## Objective 

This change aims to amend that behavior by making it explicitly known whether the schema keys are optional or required.

## Miscellaneous

* Contributes to screwdriver-cd/screwdriver#480


api/templateValidator.js
```diff
@@ -6,25 +6,36 @@ const Joi = require('joi');
...
         message: Joi.string()
+            .required()
             .label('Description of a particular validation error'),
         path: Joi.string()
+            .required()
             .label('Dot-notation path to the field that caused the validation error'),
         type: Joi.string()
+            .required()
             .label('The the Joi-type that categorizes the error')
     });
```
d2lam: I think you can use `requiredKeys()` here instead

api/templateValidator.js
```diff
@@ -6,25 +6,36 @@ const Joi = require('joi');
...
     .keys({
-        errors: Joi.array().items(TEMPLATE_ERROR),
+        errors: Joi.array()
+            .items(TEMPLATE_ERROR)
+            .required()
+            .label('Array of errors encountered while validating the given template'),
         // since a template could be parseable but invalid, the contents are unpredicatble
         template: Joi.object()
+            .required()
+            .label('The end-result of parsing the given template')
```
d2lam: same here

api/templateValidator.js
FenrirUnbound: Thanks! I didn't know that existed. Makes it sooooo much nicer 😺

api/templateValidator.js
```diff
@@ -6,25 +6,33 @@ const Joi = require('joi');
 const TEMPLATE_ERROR = Joi.object()
     .keys({
         context: Joi.object()
+            .optional()
```
d2lam: you won't need this here anymore I think

PR #113, opened 2017-03-08
feat(470): allow template field in screwdriver.yaml

d2lam:
This is for the user's side, when they specify `template` inside their `screwdriver.yaml`

This will probably be used by the `config-parser` here https://github.com/screwdriver-cd/config-parser/blob/master/lib/phase/structural.js

Related: https://github.com/screwdriver-cd/screwdriver/issues/470

### DIFF
--- a/config/regex.js
+++ b/config/regex.js
@@ -8,7 +8,9 @@ module.exports = {
     // Templates can only be named with A-Z,a-z,0-9,-,_,/
     TEMPLATE_NAME: /^[\w/-]+$/,
     // Version can only have up to 2 decimals, like 1.2.3
-    VERSION: /^(\d+\.)?(\d+\.)?(\d+)$/,
+    VERSION: /^(\d+)?(\.\d+)?(\.\d+)?$/,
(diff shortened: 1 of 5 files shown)


config/regex.js
```diff
@@ -9,6 +9,8 @@ module.exports = {
     TEMPLATE_NAME: /^[\w/-]+$/,
     // Version can only have up to 2 decimals, like 1.2.3
     VERSION: /^(\d+\.)?(\d+\.)?(\d+)$/,
```
tkyi: Should we update the regex for `VERSION` to match the end of the regex for `FULL_TEMPLATE_NAME`? Or at least make them consistent somehow
Like this: `VERSION: /^((\d+)?(\.\d+)?(\.\d+)?)$/,`

d2lam: Idk what is using `VERSION` so I don't want risk breaking whatever that's using it.

tkyi: I added `VERSION` for template validation, so it should be okay
https://github.com/screwdriver-cd/data-schema/pull/109/files

PR #118, opened 2017-04-07
fix: Add config to required keys for template

jerryzhang222:
Even though `'config.steps'` is already in the required keys for template `Joi.object`, it turns out that `'config'` itself needs to be required. Currently, this is not valid:

```yaml
name: template/test
version: 1.0.0
description: Publishes the template yaml from sd-template.yaml
maintainer: tiffanykyi@gmail.com
config:
    image: node:6
    # No steps key
```

But this is valid:

```yaml
name: template/test
version: 1.0.0
description: Publishes the template yaml from sd-template.yaml
maintainer: tiffanykyi@gmail.com
# No config key
```

### DIFF
--- a/config/template.js
+++ b/config/template.js
@@ -39,7 +39,8 @@ const SCHEMA_TEMPLATE = Joi.object()
         maintainer: TEMPLATE_MAINTAINER,
         config: Job.job
     })
-    .requiredKeys('name', 'version', 'description', 'maintainer', 'config.steps');
+    .requiredKeys('name', 'version', 'description', 'maintainer',
+        'config', 'config.image', 'config.steps');
 
 /**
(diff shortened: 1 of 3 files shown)


config/template.js
```diff
@@ -39,7 +39,7 @@ const SCHEMA_TEMPLATE = Joi.object()
         maintainer: TEMPLATE_MAINTAINER,
         config: Job.job
     })
-    .requiredKeys('name', 'version', 'description', 'maintainer', 'config.steps');
+    .requiredKeys('name', 'version', 'description', 'maintainer', 'config', 'config.steps');
```
d2lam: 👍  
can you also add config.image? to address https://github.com/screwdriver-cd/screwdriver/issues/510

PR #125, opened 2017-05-08
feat(532): Token model

ian-fox:
Add validation for tokens and their endpoints, and tests for the validation.

Currently the tokens themselves are not returned in the GET request, I figure for the one time we display them they can probably be returned to the request that was made to create the new one.


models/token.js
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @property update
+     * @type {Joi}
+     */
+    update: Joi.object(mutate(MODEL, [
+        'id'
+    ], [
+        'name',
+        'description',
+        'lastUsed'
```
FenrirUnbound: The `lastUsed` attribute is intended to describe when the token was last accessed. Why do we want the caller to be able to update it?

ian-fox: Oversight on my part

ian-fox:
@FenrirUnbound any other changes?

models/token.js
```diff
@@ -0,0 +1,114 @@
...
+    token: Joi
+        .string()
+        .token(),
+
+    id: Joi
+        .number()
+        .integer()
+        .positive(),
+
+    user: Joi
```
d2lam: `userId` would be clearer

models/token.js
```diff
@@ -0,0 +1,114 @@
...
+        .integer()
+        .positive(),
+
+    userId: Joi
+        .number()
+        .integer()
+        .positive()
+        .description('User ID'),
+
+    name: Joi
```
petey: Token names should probably be very similar to secret names.

https://github.com/screwdriver-cd/data-schema/blob/master/config/job.js#L23

ian-fox: So like SCHEMA_TOKEN, SCHEMA_ID, SCHEMA_USER_ID etc?

FenrirUnbound: Secret names are used by the build systems, so I can understand why they follow certain conventions. Token names are for the human users to better differentiate between various secrets; the system won't care what they're named. Why would we specifically want token names to be [all uppercase and without spaces](https://github.com/screwdriver-cd/data-schema/blob/master/config/regex.js#L21)

models/token.js
```diff
@@ -0,0 +1,114 @@
...
+    ])).label('Create token'),
+
+    /**
+     * Properties for token that will be passed during a UPDATE requeste
+     *
+     * @property update
+     * @type {Joi}
+     */
+    update: Joi.object(mutate(MODEL, [
+        'id'
```
minz1027: We won't need `id` here, if we are trying have a update endpoint like this: PUT /tokens/{id}.
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/secrets/update.js
The `id` shouldn't be in the payload.

ian-fox: Ooh, right. Not sure what I was thinking there, thanks!

models/token.js
```diff
@@ -0,0 +1,112 @@
...
+        .example('Mobile token'),
+
+    description: Joi
+        .string()
+        .max(256)
+        .description('Token description')
+        .example('Used to authenticate the mobile app'),
+
+    lastUsed: Joi
+        .date()
```
petey: This should probably be `.string().isoDate()` to be consistent with other timestamps.
https://github.com/screwdriver-cd/data-schema/blob/master/models/build.js#L21-L25

models/token.js
petey: ok, I'm fine with that.

models/token.js
FenrirUnbound: use `isoDate` to keep this consistent with the other time-based attributes that other model types have. 

https://github.com/screwdriver-cd/data-schema/blob/master/models/build.js#L23

ian-fox: Done!

ian-fox: Done!

ian-fox:
@petey @FenrirUnbound Made one more modification upon testing and realizing that every model needs an allKeys property. Sorry for the last minute change!

PR #133, opened 2017-05-19
fix: update fields in token schema

ian-fox:
* Get should not be an array, it will be turned into an array in the tokens.list endpoint
* Don't need a userID when creating a token, can pull that from the request authorization
* Do need a token ID to tell which token is being updated when calling token.update


models/token.js
```diff
@@ -85,7 +83,9 @@ module.exports = {
      * @property update
      * @type {Joi}
      */
-    update: Joi.object(mutate(MODEL, [], [
+    update: Joi.object(mutate(MODEL, [
+        'id'
+    ], [
```
minz1027: This is intended to be used to validate the payload of the POST tokens/{id} right?
If you look at this: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/jobs/update.js
`/{id}` is validated by validate param and the update schema is used to validate the payload. Do you want `id` inside the payload instead of part of the endpoint?

ian-fox:
Closed as authentication is being redone to trade a different access token for JWT's, rather than just using them directly

PR #134, opened 2017-05-22
fix: update fields in token schema

ian-fox:
* Store token value
* Don't need a uuid on the token any more because of storing the token value
* Get schema should only describe one token, the list endpoint will then validate an array of those.
* Don't need to pass a userId when generating a token, as we'll already have that from the authentication


models/token.js
```diff
@@ -9,12 +9,9 @@ const MODEL = {
         .integer()
         .positive(),
 
-    uuid: Joi
+    value: Joi
         .string()
```
FenrirUnbound: Should limit it to:

* The character set we would allow
* The length of the value if it's fixed-length

FenrirUnbound: Nevermind, it's going to be the sealed value

ian-fox: Done! Went with 40 characters since that's what github uses, and 36^40 is a pretty large number.

ian-fox: Oh, wait a moment, let me check if sealing preserves that stuff

ian-fox: I don't see anything in iron talking about preservation of length or the contents of the sealed string.

FenrirUnbound: The `model` will use it for saving to the datastore.

It's purely coincidence that secret values don't care about content or length. Because of that, you could use the same criteria for verifying the sealed and unsealed versions. 

However, you won't have that same guarantee with the token model. This restriction can be kept loose. Which means we can't guarantee it will be the form we say it should be (40 characters longs, etc).

PR #135, opened 2017-05-23
fix: Validate hashed token value

ian-fox:
* Need to index on the hashed value so we can look it up easily
* Not using iron means we can specify exactly what the hashed value should look like

### DIFF
--- a/models/token.js
+++ b/models/token.js
@@ -2,6 +2,12 @@
 
 const Joi = require('joi');
 const mutate = require('../lib/mutate');
+// Token length, measured in bits
+const TOKEN_LENGTH = 256;
+// Calculate the character length of the base64 string representing TOKEN_LENGTH bits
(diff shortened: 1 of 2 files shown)


models/token.js
```diff
@@ -114,5 +119,5 @@ module.exports = {
      * @property indexes
      * @type {Array}
      */
-    indexes: ['userId']
+    indexes: ['userId', 'value']
```
tkyi: `hash` instead of `value`

ian-fox: Right you are

models/token.js
```diff
@@ -2,16 +2,21 @@
 
 const Joi = require('joi');
 const mutate = require('../lib/mutate');
+const TOKEN_LENGTH = 256;
```
minz1027: Can you provide extra comment above this line about why the token_length is 256 and why the HASH_LENGTH is 256 / 24?

PR #137, opened 2017-06-02
fix(tokens): Sequelize didn't like null values

ian-fox:
The datastore wasn't validating tokens that hadn't been used, because the `.allow(null)` was being ignored by the datastore-sequelize plugin. Now using empty strings instead of `null`.


tkyi:
Wondering if it would be better to adjust the sequelize plugin to handle null cases better.

PR #139, opened 2017-06-05
fix(tokens): Encode tokens with URI-safe encoding

ian-fox:
* tokens were previously stored in base64 encoding, which is not URI-safe
* switched encoding to hexadecimal

### DIFF
--- a/models/token.js
+++ b/models/token.js
@@ -4,10 +4,8 @@ const Joi = require('joi');
 const mutate = require('../lib/mutate');
 // Token length, measured in bits
 const TOKEN_LENGTH = 256;
-// Calculate the character length of the base64 string representing TOKEN_LENGTH bits
-// Each base64 character represents 6 bits of data, and strings are padded to be a
-// multiple of 4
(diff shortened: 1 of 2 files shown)


FenrirUnbound:
What is wrong with storing the tokens in base64 instead of in hexadecimal? 

Are we planning on using the token in a URL?

ian-fox:
The login route takes tokens through the URL. `encodeURIComponent` could work if people try to hit the login route directly, trying to figure out now if whatever it is that redirects them to the login route will automatically put an access token in the header into the URL, and if so if it calls encodeURIComponent.

ian-fox:
Closing because more research is required

ian-fox:
The endpoint for authentication can take tokens through the query string or through a `bearer` header. We're already using `bearer` for JWT's. We could pattern match things to see if they're one or the other I guess, but that seems suspect.

Assuming we then go with URL-encoded tokens, I feel like making people have to call `encodeURIComponent` (or whatever the equivalent in their language of choice is) is bad user experience.

ian-fox:
Closing as more investigation is needed into whether we can use the `authorization` header with both the hapi-bearer and hapi-jwt plugins.

PR #140, opened 2017-06-06
fix(tokens): Use url-safe base64 encoding

ian-fox:
* the default base64 encoding uses `+`, `/`, and `=`
* replace those with `_`, `-`, and `~`

### DIFF
--- a/models/token.js
+++ b/models/token.js
@@ -17,7 +17,7 @@ const MODEL = {
 
     hash: Joi
         .string()
-        .base64()
+        .regex(/[a-zA-Z0-9_-]+~/)
         .length(HASH_LENGTH)
         .description('Hashed token value'),
 
--- a/test/data/token.yaml
+++ b/test/data/token.yaml
@@ -1,6 +1,6 @@
 # Base Token Example
 userId: 1234
-hash: 'aHashedTokenValueMustBe44CharactersLong+++++'
+hash: 'aHashedTokenValueMustBe44CharactersLong_-_-~'
 id: 1111
 name: 'Auth token'
 description: 'A token for authentication'


models/token.js
```diff
@@ -17,7 +17,7 @@ const MODEL = {
 
     hash: Joi
         .string()
-        .base64()
+        .regex(/[a-zA-Z0-9_-]+~/)
```
FenrirUnbound: Can we not use periods `.` either?

ian-fox: We could, if there's some reason those are preferable. There are only three characters we need to replace though.

FenrirUnbound: I'd have to look it up again, but I thought there were 3 characters that weren't URI safe: `/`, `+`, and `=`. I've seen implementations where the `+` gets converted over to a `.`.

I guess it depends on the base64 library and what does underneath the hood

FenrirUnbound: [Looking at `base64url`](https://github.com/brianloveswords/base64url/blob/master/src/base64url.ts#L27-L29), it translates the `=` into nothing. So your character-set choice seems valid. 

I'd recommend a comment here that it's based on this library's translation table, in case someone in the future decides to use a different library.

ian-fox: I feel like it might be simpler just to define a function in the one file that deals with string encodings (models/generateToken.js) that runs the replace, rather than bringing in a dependency. All that needs to be done is run `string.replace` a few times. I could be wrong on that though.

FenrirUnbound: So now you have to know that the characters in this `regex` are dependent on an implementation in a library that's not referenced anywhere in this file. It'll just suck for the next developer that has to debug this.

ian-fox: Ah, hadn't thought of it that way. I'll move to the library then.

PR #145, opened 2017-06-12
feat(580): support for annotations

r3rastogi:
Annotations are a grouping of key-value pairs and can be either pipeline- or job-level specifications.

More details available at: screwdriver-cd/screwdriver#580

### DIFF
--- a/api/validator.js
+++ b/api/validator.js
@@ -1,8 +1,9 @@
 'use strict';
 
+const Annotations = require('../config/annotations');
+const Job = require('../config/job');
 const Joi = require('joi');
 const Regex = require('../config/regex');
-const Job = require('../config/job');
 const Workflow = require('../config/workflow');
 
 const SCHEMA_JOB_COMMAND = Joi.object()
@@ -20,9 +21,10 @@ const SCHEMA_JOB_COMMANDS = Joi.array()
 
 const SCHEMA_JOB_PERMUTATION = Joi.object()
     .keys({
-        image: Job.image,
+        annotations: Annotations.annotations,
         commands: SCHEMA_JOB_COMMANDS,
         environment: Job.environment,
+        image: Job.image,
         secrets: Job.secrets,
         settings: Job.settings
     }).label('Job permutation');
@@ -42,9 +44,10 @@ const SCHEMA_JOBS = Joi.object()
 
 const SCHEMA_OUTPUT = Joi.object()
     .keys({
+        annotations: Annotations.annotations,
+        errors: Joi.array().items(Joi.string()).optional(),
         jobs: SCHEMA_JOBS,
-        workflow: Workflow.workflow,
-        errors: Joi.array().items(Joi.string()).optional()
+        workflow: Workflow.workflow
     })
     .label('Execution information');
 
--- a/config/annotations.js
+++ b/config/annotations.js
@@ -1,0 +1,14 @@
+'use strict';
+
+const Joi = require('joi');
+
+// List of data, eventually key-value pairs
+const SCHEMA_ANNOTATIONS = Joi.object().optional();
+
+/**
+ * The definition of the annotations pieces
+ * @type {Object}
+ */
+module.exports = {
+    annotations: SCHEMA_ANNOTATIONS
+};
--- a/config/job.js
+++ b/config/job.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const Annotations = require('./annotations');
 const Joi = require('joi');
 const Regex = require('./regex');
 
@@ -65,12 +66,13 @@ const SCHEMA_SETTINGS = Joi.object().optional();
 const SCHEMA_TEMPLATE = Joi.string().regex(Regex.FULL_TEMPLATE_NAME);
 const SCHEMA_JOB = Joi.object()
     .keys({
-        steps: SCHEMA_STEPS,
+        annotations: Annotations.annotations,
         environment: SCHEMA_ENVIRONMENT,
-        matrix: SCHEMA_MATRIX,
         image: SCHEMA_IMAGE,
+        matrix: SCHEMA_MATRIX,
         secrets: SCHEMA_SECRETS,
         settings: SCHEMA_SETTINGS,
+        steps: SCHEMA_STEPS,
         template: SCHEMA_TEMPLATE
     })
     .default({});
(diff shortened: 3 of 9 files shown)


test/data/config.annotations.yaml
```diff
@@ -0,0 +1,2 @@
+- envA/foo: true
```
tkyi: Can you put examples from the issue in here?
https://github.com/screwdriver-cd/screwdriver/issues/580

config/annotations.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const Joi = require('joi');
+
+// List of data, eventually key-value pairs
+const SCHEMA_ANNOTATIONS = Joi.array()
```
stjohnjohnson: Isn't this just an `Joi.object().optional()`?  I don't think it's an array.  Order doesn't matter.

config/job.js
```diff
@@ -65,12 +66,13 @@ const SCHEMA_SETTINGS = Joi.object().optional();
 const SCHEMA_TEMPLATE = Joi.string().regex(Regex.FULL_TEMPLATE_NAME);
 const SCHEMA_JOB = Joi.object()
     .keys({
-        steps: SCHEMA_STEPS,
+        annotations,
```
stjohnjohnson: Does this become `annotations: {annotations: obj}`

models/job.js
```diff
@@ -41,7 +42,10 @@ const MODEL = {
         .boolean()
         .description('Flag if the job is archived')
         .example(true)
-        .default(false)
+        .default(false),
+
+    annotations: Annotations.annotations
```
d2lam: Why would we add it to model? It shouldn't be a field in the database

models/pipeline.js
```diff
@@ -39,7 +40,10 @@ const MODEL = {
         .example({ myself: true }),
 
     workflow: Workflow.workflow
-        .description('Current workflow of the pipeline')
+        .description('Current workflow of the pipeline'),
+
+    annotations: Annotations.annotations
```
d2lam: Same comment here. I don't think we need to add it to models?

api/validator.js
```diff
@@ -44,7 +45,8 @@ const SCHEMA_OUTPUT = Joi.object()
     .keys({
         jobs: SCHEMA_JOBS,
         workflow: Workflow.workflow,
-        errors: Joi.array().items(Joi.string()).optional()
+        errors: Joi.array().items(Joi.string()).optional(),
+        annotations: Annotations.annotations
```
d2lam: I don't think this should be here. I think it should be inside `SCHEMA_JOB_PERMUTATION` instead.

r3rastogi: What does `SCHEMA_JOB_PERMUTATION` do?

models/job.js
r3rastogi: Ahh I didn't realize adding something to models added it to the database. Will remove.

config/annotations.js
r3rastogi: Hm my understanding was that all the annotations were a list of key-value pairs aka an array. But no worries - will change to an object.

test/data/config.annotations.yaml
r3rastogi: Sure

config/job.js
r3rastogi: I'm not sure what you mean here

models/pipeline.js
r3rastogi: Will remove

api/validator.js
d2lam: It's the one after merged. So `config-parser` will merge share & job level config in the correct order, then put into job permutations.

config/job.js
tkyi: I think he means you'll want to have the key "annotations" followed by the actual Joi validation object for annotations

minz1027: @r3rastogi I think what he meant is this:
```
> var annotations = {"annotations": true}
undefined
> var obj = {annotations}
undefined
> obj
{ annotations: { annotations: true } }
> 
```
But we actually do not want the extra layer.

api/validator.js
stjohnjohnson: There should be a pipeline level annotations.

config/job.js
```diff
@@ -2,6 +2,7 @@
 
 const Joi = require('joi');
 const Regex = require('./regex');
+const annotations = require('./annotations');
```
tkyi: Maybe rename to `Annotations` for consistency

config/job.js
```diff
@@ -80,14 +82,15 @@ const SCHEMA_JOB = Joi.object()
...
-    matrix: SCHEMA_MATRIX,
-    secret: SCHEMA_SECRET,
-    secrets: SCHEMA_SECRETS,
-    steps: SCHEMA_STEPS,
-    step: SCHEMA_STEP,
+    annotations,
     environment: SCHEMA_ENVIRONMENT,
     image: SCHEMA_IMAGE,
     job: SCHEMA_JOB,
+    matrix: SCHEMA_MATRIX,
```
tkyi: Yay for alphabetization! 🎉

config/job.js
r3rastogi: Oops, yes, my bad

config/job.js
r3rastogi: Ohh yes that makes sense! Fixing now.

api/validator.js
d2lam: @r3rastogi  Sorry, I thought you wouldn't need in pipeline. Can you add it back? :p 
You still need it in `SCHEMA_JOB_PERMUTATION` though.

r3rastogi: @d2lam @stjohnjohnson updated. It's now in both `SCHEMA_JOB_PERMUTATION` (for job-level) and `SCHEMA_OUTPUT` (for pipeline-level)

config/annotations.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const Joi = require('joi');
+
+// List of data, eventually key-value pairs
+const SCHEMA_ANNOTATIONS = Joi.object().optional();
```
ian-fox: I can see why it would be from a code-reuse point of view, but it seems a bit odd to me that the `optional` flag is on the annotations schema itself, rather than the places where annotations are optional.

r3rastogi: I had based `annotations` off of how `workflow` was implemented. If there is a preference to define everything locally instead of its own class, I'm amenable to that.

ian-fox: I'm not actually sure which would be preferable, just something that struck me as slightly odd.

PR #146, opened 2017-06-13
fix(604): Add annotations to executor start

tkyi:
Adding annotations to the start schema for executors. This will allow people to send annotations to specify which executor to use when starting a build.

Example annotations yaml segment:
```
annotations:
    beta.screwdriver.cd/executor: k8s
```

Related to https://github.com/screwdriver-cd/screwdriver/issues/604

### DIFF
--- a/plugins/executor.js
+++ b/plugins/executor.js
@@ -1,9 +1,11 @@
 'use strict';
 
+const Annotations = require('../config/annotations');
 const Joi = require('joi');
 const models = require('../models');
 const buildId = Joi.reach(models.build.base, 'id').required();
 const SCHEMA_START = Joi.object().keys({
+    annotations: Annotations.annotations,
     buildId,
     container: Joi.reach(models.build.base, 'container').required(),
     apiUri: Joi.string().uri().required()
--- a/test/data/executor.startNoAnnotations.yaml
+++ b/test/data/executor.startNoAnnotations.yaml
@@ -1,5 +1,3 @@
-annotations:
-    beta.screwdriver.cd/executor: k8s
 buildId: 8609
 container: node:4
 apiUri: http://localhost:8080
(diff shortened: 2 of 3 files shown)


plugins/executor.js
```diff
@@ -1,9 +1,11 @@
 'use strict';
 
+const Annotations = require('../config/annotations');
 const Joi = require('joi');
 const models = require('../models');
 const buildId = Joi.reach(models.build.base, 'id').required();
 const SCHEMA_START = Joi.object().keys({
+    annotations: Annotations.annotations,
```
d2lam: When we get to the `start` function, isn't it already **inside** the executor plugin? I'm a little confused on how this would work. Could you give some details on how this is going to be used?

FenrirUnbound: Is the issue not descriptive enough? https://github.com/screwdriver-cd/screwdriver/issues/604

test/data/executor.start.yaml
```diff
@@ -1,3 +1,5 @@
+annotations:
+    beta.screwdriver.cd/executor: k8s
```
minz1027: Are we supposed to have some rules about the key of the annotations? We just assume the `executor-router`? knows how to parse it?

tkyi: I think we'll have a list of recognized annotations much like Kubernetes does:
https://kubernetes.io/docs/reference/labels-annotations-taints/

FenrirUnbound: > Are we supposed to have some rules about the key of the annotations? 

Depends on the work in #595, specifically the "Consume annotations to choose executor resource"

> We just assume the executor-router? knows how to parse it? This is just an example or we are sure we will use this format in executor-router?

My counter question would be: how does anyone know what value goes [in the config](https://github.com/screwdriver-cd/screwdriver/blob/master/config/default.yaml#L89)? The annotation value would have to match the module's name, like how we're currently doing it [in bin/server](https://github.com/screwdriver-cd/screwdriver/blob/master/bin/server#L37)

minz1027: We have documentation for the executor plugin in the our guide right? I think @tkyi makes a good point about how to document it. Once we have figured out the rules to "Consume annotations to choose executor resource", we can document it.

PR #149, opened 2017-06-20
feat(423): Allow descriptions in yaml file

joelseq:
- Add a schema for description
- Add a test data file (config.job.description.yaml)
- Add a test to validate description

PR #151, opened 2017-06-27
fix(423): Add job descriptions to validator

joelseq:
- Add SCHEMA_DESCRIPTION to SCHEMA_JOB_PERMUTATION and SCHEMA_OUTPUT
- Modify validator.input.yaml and validator.output.yaml to make sure
  description and annotations work (related to annotations PR https://github.com/screwdriver-cd/data-schema/pull/145).

Part of #149

### DIFF
--- a/api/validator.js
+++ b/api/validator.js
@@ -23,6 +23,7 @@ const SCHEMA_JOB_PERMUTATION = Joi.object()
     .keys({
         annotations: Annotations.annotations,
         commands: SCHEMA_JOB_COMMANDS,
+        description: Job.description,
         environment: Job.environment,
         image: Job.image,
         secrets: Job.secrets,
--- a/test/data/validator.input.yaml
+++ b/test/data/validator.input.yaml
@@ -1,6 +1,9 @@
 yaml: |
   jobs:
       main:
+          annotations:
+              beta.screwdriver.cd/auto_pr_builds: fork-only
+          description: "This is a description"
           image: node:{{NODE_VERSION}}
           matrix:
               NODE_VERSION: [4,5,6]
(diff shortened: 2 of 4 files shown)


test/data/validator.output.yaml
```diff
@@ -1,5 +1,8 @@
 jobs:
   main:
+  - annotations:
```
d2lam: I'm kinda confused. Shouldn't this be in the same level as `settings` and `environments`?

api/validator.js
```diff
@@ -44,7 +44,8 @@ const SCHEMA_JOBS = Joi.object()
 
 const SCHEMA_OUTPUT = Joi.object()
     .keys({
-        annotations: Annotations.annotations,
+        annotations: Job.annotations,
+        description: Job.description,
```
minz1027: Why we need it here? Should be inside the job permutation already right?

test/data/validator.input.yaml
```diff
@@ -1,6 +1,9 @@
 yaml: |
   jobs:
       main:
+          annotations:
+              beta.screwdriver.cd/auto_pr_builds: fork-only
+          description: "This is a description"
```
d2lam: I'm kinda confused. Shouldn't this be in the same level as settings and environments?

test/data/validator.output.yaml
```diff
@@ -1,5 +1,8 @@
...
+      beta.screwdriver.cd/auto_pr_builds: fork-only
+  - description: "This is a description"
```
d2lam: I'm kinda confused. Shouldn't this be in the same level as settings and environments?

PR #152, opened 2017-07-05
feat(365): Add scmContext to models

s-yoshika:
This PR add `scmContext` to models.It is a change for the support multiple SCMs.
`scmContext` determines that an user is currently logged into by which scm.

### DIFF
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -27,6 +27,11 @@ const MODEL = {
         .description('Unique identifier for the application')
         .example('github.com:123456:master'),
 
+    scmContext: Joi
+        .string().max(128)
+        .description('The SCM in which the repository exists')
+        .example('github.com'),
+
     scmRepo: Scm.repo,
 
     createTime: Joi
@@ -62,7 +67,7 @@ module.exports = {
      * @type {Joi}
      */
     get: Joi.object(mutate(MODEL, [
-        'id', 'scmUri', 'createTime', 'admins'
+        'id', 'scmUri', 'scmContext', 'createTime', 'admins'
     ], [
         'workflow', 'scmRepo', 'annotations'
     ])).label('Get Pipeline'),
--- a/models/user.js
+++ b/models/user.js
@@ -17,7 +17,13 @@ const MODEL = {
 
     token: Joi
         .string()
-        .description('Github token')
+        .description('Github token'),
+
+    scmContext: Joi
+        .string()
+        .max(128)
+        .description('The SCM to which the user belongs')
+        .example('github.com')
 };
 
 module.exports = {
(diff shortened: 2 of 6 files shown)


tk3fftk:
Related: https://github.com/screwdriver-cd/screwdriver/issues/365

models/pipeline.js
```diff
@@ -27,6 +27,11 @@ const MODEL = {
         .description('Unique identifier for the application')
         .example('github.com:123456:master'),
```
tkyi: Could we just use the first value before the first colon in the `scmUri`? Would this value ever be different from `scmContext`?

s-yoshika: Yes, we can use `scmUri` instead of `scmContext`, but I think that `scmContext` is effective when we want to filter pipelines by the scm  by which user log-ined.

tkyi: Oh I see. Yeah it makes sense to have `scmContext` for pipeline filtering. 👍

minz1027: Just to remind, currently we cannot add a new column to existing tables automatically. Probably need to add that column manually if we want to plug it in now.

Related to this issue:
https://github.com/screwdriver-cd/screwdriver/issues/633

s-yoshika: Thanks for helpfull information!

models/pipeline.js
```diff
@@ -62,7 +67,7 @@ module.exports = {
      * @type {Joi}
      */
     get: Joi.object(mutate(MODEL, [
-        'id', 'scmUri', 'createTime', 'admins'
+        'id', 'scmUri', 'scmContext', 'createTime', 'admins'
```
d2lam: Does this mean it will break once we pull in this new data-schema? I'm wondering if we should put this as `optional` field first.

minz1027: Should we wait till the sequelize problem is fixed and then merge this? I think they want the scmContext to be required.

d2lam: But that means it will be blocked by the sequelize team. We don't know when they would publish a new version. Would there be any harm in leaving it as `optional` then `required` later?

s-yoshika: We plans to send a pull-request for api which requires the user schema has `scmContext`.
So, in users schema, `scmContext` is required.

To migrate these tables, I submit the comment to the issue.
https://github.com/screwdriver-cd/screwdriver/issues/365#issuecomment-313570292

d2lam: Okay I'm fine with this. We just need to do the migration right after this is pulled in.

d2lam:
@tk3fftk  Can you do a major version bump since this is breaking?

tk3fftk:
@s-yoshika 
> Can you do a major version bump since this is breaking?

Please do it🚀

d2lam:
@tk3fftk @s-yoshika I just remembered that if we bump major version, we would need to bump it in all the repos since there is a check in the API that checks if all repos have the correct major version. It might be too troublesome to do that, so I'm ok with bumping minor.

PR #153, opened 2017-07-07
feat(523): Add collection schemas and tests

joelseq:
# Context

Schema validation for Collections

# Objective

This PR adds a schema for Collections along with tests to be used for the creation of a CRUD API for collections

# References

### DIFF
--- a/models/collection.js
+++ b/models/collection.js
@@ -1,0 +1,114 @@
+'use strict';
+
+const Joi = require('joi');
+const mutate = require('../lib/mutate');
+
+const MODEL = {
+    id: Joi
+        .number()
+        .integer()
+        .positive(),
+
+    userId: Joi
+        .number()
+        .integer()
+        .positive()
+        .description('User ID'),
+
+    name: Joi
+        .string()
+        .max(128)
+        .description('Collection name')
+        .example('Favorites'),
+
+    description: Joi
+        .string()
+        .max(256)
+        .description('Collection description')
+        .example('List of my favorite pipelines'),
+
+    pipelineIds: Joi
+        .array()
+        .items(Joi.number().integer().positive())
+};
+
+module.exports = {
+    /**
+     * All the available properties of Collection
+     *
+     * @property base
+     * @type {Joi}
+     */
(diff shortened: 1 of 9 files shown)


models/collection.js
```diff
@@ -0,0 +1,113 @@
...
+        'description'
+    ])).label('Get collecion'),
+
+    /**
+     * Properties for Collection that will be passed during a CREATE request
+     *
+     * @property create
+     * @type {Joi}
+     */
+    create: Joi.object(mutate(MODEL, [
```
ian-fox: Do we want the ability to make a new collection with something already in it?

When you're making a playlist in some sort of music player, for example, you can generally select a song, hit "add to playlist," and then choose "make a new one" and put in a name and it'll be created with that song already populated.

models/collection.js
```diff
@@ -0,0 +1,113 @@
...
+        .string()
+        .max(128)
+        .description('Collection name')
+        .example('Favorites'),
+
+    description: Joi
+        .string()
+        .max(256)
+        .description('Collection description')
+        .example('List of my favorite collections'),
```
d2lam: favorite pipelines?

models/collection.js
```diff
@@ -0,0 +1,113 @@
...
+     * @property get
+     * @type {Joi}
+     */
+    get: Joi.object(mutate(MODEL, [
+        'id',
+        'name',
+        'pipelineIds'
+    ], [
+        'description'
+    ])).label('Get collecion'),
```
tkyi: collection*

models/collection.js
tkyi: I think it depends on our mechanism for creating Collections. Would you ever just type in `pipelineIds` you'd want to add, or is it always going to be some drag and drop sort of interaction?

joelseq: I was initially thinking that creating a pipeline would just have like a name and an optional description and then pipelines could be added through update requests. But I can see how in an API standpoint it would be nice to support that functionality. I can see it being useful if we are going to go with the fork model for modifying someone else's collection.

PR #157, opened 2017-07-11
feat: add template tag

d2lam:
Add a Template Tag model, which consists of name, tag, and version. 

Example: `mytemplate@1.2.0` can be tagged as `stable` and `latest`

In the templateTags table, there will be a record that looks like 

| id | name | tag | version | 
|---|---|---|---|
| 1 | mytemplate | stable | 1.2.0 | 
| 2 | mytemplate | latest | 1.2.0 | 

If they want to change `stable` to `1.1.0`, then the table will look like 

| id | name | tag | version | 
|---|---|---|---|
| 1 | mytemplate | stable | 1.1.0 | 
| 2 | mytemplate | latest | 1.2.0 | 

Related: https://github.com/screwdriver-cd/screwdriver/issues/616

### DIFF
--- a/models/template.js
+++ b/models/template.js
@@ -55,7 +55,7 @@ module.exports = {
     ], ['labels'])).label('Create Template'),
 
     /**
-     * Properties for template that will be passed during a UPDATE requeste
+     * Properties for template that will be passed during a UPDATE request
      *
      * @property update
      * @type {Joi}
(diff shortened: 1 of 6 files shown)


models/template.js
```diff
@@ -55,7 +55,7 @@ module.exports = {
     ], ['labels'])).label('Create Template'),
 
     /**
-     * Properties for template that will be passed during a UPDATE requeste
+     * Properties for template that will be passed during a UPDATE request
```
tkyi: 👍

PR #161, opened 2017-07-14
fix(collections): Change the GET schema to include pipelines

joelseq:
# Context

Currently, the GET schema is designed to only return an array of
pipeline IDs but it would make more sense if the GET for a collection
returned an array of the pipeline objects.

# Objective

This PR changes the GET schema of collection so that an array of pipeline objects are sent back when a collection is requested instead of just the IDs for the pipelines in that collection

# References

### DIFF
--- a/models/collection.js
+++ b/models/collection.js
@@ -3,6 +3,8 @@
 const Joi = require('joi');
 const mutate = require('../lib/mutate');
 
+const PIPELINE_MODEL = require('./pipeline').get;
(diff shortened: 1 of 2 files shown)


models/collection.js
```diff
@@ -2,6 +2,9 @@
 
 const Joi = require('joi');
 const mutate = require('../lib/mutate');
+const PIPELINE_MODEL = require('./pipeline').get;
```
r3rastogi: non-blocking, but could you alphabetize the consts and not include extra newlines please :)

PR #162, opened 2017-07-19
fix(collections): Add list schema for collections

joelseq:
# Context

Due to the difference in the structure of a collection object in the GET and LIST API responses, we need a separate schema for LIST. In the LIST request, the UI will primarily require id, name, and description for each of the user's collections. The GET request will populate all the pipelineIds with the pipeline objects instead.

# Objective

This PR adds a schema for the LIST request along with a unit test.

# References 

### DIFF
--- a/models/collection.js
+++ b/models/collection.js
@@ -59,6 +59,19 @@ module.exports = {
     ])).label('Get collection'),
 
     /**
+     * Properties for collections that will come back during a LIST request.
+     * The LIST request will list all of the requesting user's collections.
+     */
+    list: Joi.array().items(Joi.object(mutate(MODEL, [
+        'id',
+        'name',
+        'pipelineIds',
+        'userId'
+    ], [
+        'description'
+    ]))).label('List collections for requesting user'),
+
+    /**
      * Properties for Collection that will be passed during a CREATE request
      *
      * @property create
--- a/test/data/collection.list.yaml
+++ b/test/data/collection.list.yaml
@@ -1,0 +1,15 @@
+- id: 12
+  name: 'collection1'
+  description: 'pipelines for collection 1'
+  pipelineIds: [11, 12, 13]
+  userId: 123
+- id: 13
+  name: 'collection2'
+  description: 'pipelines for collection 2'
+  pipelineIds: [14, 15, 16]
+  userId: 123
+- id: 14
+  name: 'collection3'
+  description: 'pipelines for collection 3'
+  pipelineIds: [17, 18, 19]
+  userId: 123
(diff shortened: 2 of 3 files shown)


models/collection.js
```diff
@@ -59,6 +59,17 @@ module.exports = {
     ])).label('Get collection'),
 
     /**
+     * Properties fo collections that will come back during a LIST request
```
ian-fox: Typo in "fo"

ian-fox:
Also, not sure if we actually need a list schema here. In [secrets](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/listSecrets.js#L6), for instance, we say "the list schema is just an array of objects that all match the get schema"

ian-fox:
If the issue is that you want IDs in the list, it may be better just to add `id` to the get schema. That makes it consistent with other things, again like [secrets](https://github.com/screwdriver-cd/data-schema/blob/master/models/secret.js#L48) or [pipelines](https://github.com/screwdriver-cd/data-schema/blob/master/models/pipeline.js#L65)

joelseq:
@ian-fox well the issue is more about populating pipelines in. In the GET request, I actually populate the pipelines in which is why the schema has pipelines and not pipelineIds. So an example response of a **GET** request for a particular collection would look like:

```json
{
    "id": 12,
    "name": "screwdriver",
    "description": "pipelines for screwdriver",
    "pipelines": [
        {
            "id": 123,
            "scmUri": "github.com:12345:branchName",
            "createTime": "2038-01-19T03:14:08.131Z",
            "admins": {
                "stjohn": true
            }
        },
        {
            "id": 124,
            "scmUri": "github.com:12345:branchName",
            "createTime": "2038-01-19T03:14:08.131Z",
            "admins": {
                "tkyi": true
            }
        },
        {
            "id": 125,
            "scmUri": "github.com:12345:branchName",
            "createTime": "2038-01-19T03:14:08.131Z",
            "admins": {
                "d2lam": true
            }
        }
    ]
}
```

But the **LIST** request to get the list of collections for a user would look like:

```json
[
    {
        "id": 12,
        "name": "collection1",
        "description": "pipelines for collection 1",
        "pipelineIds": [123, 124, 125]
    },
    {
        "id": 13,
        "name": "collection2",
        "description": "pipelines for collection 2",
        "pipelineIds": [123, 124, 125]
    },
    {
        "id": 14,
        "name": "collection3",
        "description": "pipelines for collection 3",
        "pipelineIds": [123, 124, 125]
    }
]
```

So in this case, I cannot really do `Joi.array.items(schema.models.collection.get)` for the LIST request. I could make both `pipelines` and `pipelineIds` optional in the GET schema in which case it will work but I don't know if that would be optimal?

PR #167, opened 2017-07-25
[DNMy] feat(365) - 1: Add scmContext to models #152

catto:
Reverts screwdriver-cd/data-schema#160

We're going to complete implementing the multiple scms feature. screwdriver-cd/screwdriver#365
This PR is the first one you can start merging. I'll let you know when it is ready to review and merge.

### Changes since last PR
Added `scmContext` key to test data for collection due to schema changes.

### DIFF
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -30,7 +30,7 @@ const MODEL = {
     scmContext: Joi
         .string().max(128)
         .description('The SCM in which the repository exists')
-        .example('github.com'),
+        .example('github:github.com'),
 
     scmRepo: Scm.repo,
 
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -27,6 +27,11 @@ const MODEL = {
         .description('Unique identifier for the application')
         .example('github.com:123456:master'),
 
+    scmContext: Joi
+        .string().max(128)
+        .description('The SCM in which the repository exists')
+        .example('github.com'),
+
     scmRepo: Scm.repo,
 
     createTime: Joi
@@ -62,7 +67,7 @@ module.exports = {
      * @type {Joi}
      */
     get: Joi.object(mutate(MODEL, [
-        'id', 'scmUri', 'createTime', 'admins'
+        'id', 'scmUri', 'scmContext', 'createTime', 'admins'
     ], [
         'workflow', 'scmRepo', 'annotations'
     ])).label('Get Pipeline'),
--- a/models/user.js
+++ b/models/user.js
@@ -17,7 +17,13 @@ const MODEL = {
 
     token: Joi
         .string()
-        .description('Github token')
+        .description('Github token'),
+
+    scmContext: Joi
+        .string()
+        .max(128)
+        .description('The SCM to which the user belongs')
+        .example('github.com')
 };
 
 module.exports = {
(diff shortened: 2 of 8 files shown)


ian-fox:
I might still be missing something: why do some of the test data have `github:github.com` and some have just `github.com`?

minz1027:
@ian-fox That's for the GHE host, enterprise github.  https://github.com/kumada626/scm-github/blob/01fd0ad1692d53400e53bac9364e5d7d4c2fe585/index.js#L788

Looks like for github, it's ['github.com'] or ['github: ghehost'] if you have ghehost.

But for bitbucket, the default one is ['bitbucket: bitbucket.org']? It's not consistent.
https://github.com/sakka2/scm-bitbucket/blob/a944ffe2c8c2744fc8a9403b62e5a29718d85e9b/index.js#L763

Can we document the `scmContext` interface clearly somewhere so that when we review we have better context?

catto:
@ian-fox Er they include old test data because this is based on the revert commit. I'll fix them so that they would be easy for us to understand.

d2lam:
@ian-fox  @catto  
Why don't we make it consistent: `github:github.com` and  `github:mygithubenterprise.com`
So the format will be consistent as `scm:host`

catto:
@d2lam Some scms in the future might not be distinguished by only hostname. I think it is better to let scm module handle the format of scm context.
But I still agree with you about generating`github:github.com` and `github:mygithubenterprise.com`. Let's fix scm-module to generate `github:github.com` when gheHost is not set.

@kumada626 🙇

d2lam:
Should we bump major?

petey:
From the perspective of tests as documentation for code, I feel like we are overly tying our test data to look like this works only github.com or GHE variants. It might be worthwhile to instead have the tests reproduce the interface of the scmContext without tying to a specific property. e.g. scmContext: `scmType:scm.host`, or scmUri: `scmhost:1234:branch`

d2lam:
I agree with Petey but maybe we can do that in another PR, since `scmUri` is not part of this PR.
But I think we should at least change `github.com` to `github:github.com`, since that is the current implementation of `scm-github`.

models/pipeline.js
```diff
@@ -27,6 +27,11 @@ const MODEL = {
         .description('Unique identifier for the application')
         .example('github.com:123456:master'),
 
+    scmContext: Joi
+        .string().max(128)
+        .description('The SCM in which the repository exists')
+        .example('github.com'),
```
petey: All the `scmContext` examples are incorrect, when compared to the tests.

catto: @petey I've fixed them 🙇  also some test data

models/user.js
```diff
@@ -17,7 +17,13 @@ const MODEL = {
 
     token: Joi
         .string()
-        .description('Github token')
+        .description('Github token'),
+
+    scmContext: Joi
+        .string()
```
petey: We may want to add a `.regex(...)` to this at some point.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "screwdriver-data-schema",
-  "version": "16.0.0",
+  "version": "16.1.0",
```
tkyi: Can we bump major for this one since scmContext is required in a `get` for pipelines?

PR #168, opened 2017-08-07
feat: add queue job config validation schema

minz1027:
Add queue job config validation schema

`executor-queue` will use this schema to validate the job config before it sends it to the queue.

Related to https://github.com/screwdriver-cd/screwdriver/issues/431


test/data/queue.start.yaml
```diff
@@ -0,0 +1,11 @@
+buildConfig:
+    buildId: 8609
+    container: node:4
+    apiUri: http://localhost:8080
+    token: eyJhbGciOiJIUziIsInR5cCI6IkpXVCJ9.eyJ1c2Vybam9obmpvaG5zb24iLCJzY29wZSI6WyJ1c2VyIl0sImlhdCI6MTQ3MDI2NjAwMywiZXhwIjoxNDcwMzA5MjAzfQ.CHaHW2-0wALpWaS4UTXQmSAn_ekMBml-ENEnGhw-9SE
+executor:
```
r3rastogi: Do we not need an `annotations` field to specify the type of executor anymore?

minz1027: The routing logic should be taken care of by the executor-queue. When the executor-queue enqueues a job, it should know what executor to be picked already.

The executor-queue will check the annotation to pick the right build executor.

PR #172, opened 2017-08-19
feat: Allow Multiple-Documents

stjohnjohnson:
This adds an optional field in the base config to allow us to hint about
what version the config is.


config/base.js
```diff
@@ -17,6 +17,7 @@ const SCHEMA_JOBS = Joi.object()
 const SCHEMA_SHARED = Job.job;
 const SCHEMA_CONFIG = Joi.object()
     .keys({
+        version: Joi.number().integer(),
```
tkyi: Should there be a `.min(1).max(50)`

PR #178, opened 2017-08-26
feat(collections): Add pr info to pipelines in collections get schema

joelseq:
## Context

For the user dashboards, we want to display information about the number of PRs open and failing for each pipeline in the collection.

## Objective

This PR adds pr info to each pipeline in the GET for a collection

## References

Issue: [screwdriver-cd/screwdriver#523](https://github.com/screwdriver-cd/screwdriver/issues/523)

### DIFF
--- a/models/collection.js
+++ b/models/collection.js
@@ -5,8 +5,13 @@ const mutate = require('../lib/mutate');
 
 const BUILD_MODEL = require('./build').get;
 const PIPELINE_MODEL = require('./pipeline').get;
+const PRS = {
+    open: Joi.number().integer().min(0),
+    failing: Joi.number().integer().min(0)
+};
 const PIPELINE_OBJECT = PIPELINE_MODEL.keys({
-    lastBuilds: Joi.array().items(BUILD_MODEL).optional()
+    lastBuilds: Joi.array().items(BUILD_MODEL).optional(),
+    prs: Joi.object(PRS).optional()
 });
 const PIPELINES_MODEL = Joi.array().items(PIPELINE_OBJECT);
 const MODEL = {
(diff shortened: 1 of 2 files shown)


models/collection.js
```diff
@@ -5,8 +5,13 @@ const mutate = require('../lib/mutate');
 
 const BUILD_MODEL = require('./build').get;
 const PIPELINE_MODEL = require('./pipeline').get;
+const PRS = {
```
d2lam: We just want the numbers? Or we want the last build of opened PRs?

joelseq: According to the [mock](https://user-images.githubusercontent.com/78533/27356398-2694102e-55c3-11e7-877f-ec96d074bbf0.png) we just want the numbers

models/collection.js
```diff
@@ -5,8 +5,13 @@ const mutate = require('../lib/mutate');
...
+    open: Joi.number().integer(),
```
tkyi: `.min(0)`?

joelseq: I was contemplating putting `.positive()` but wasn't sure if 0 would count or not so I just didn't add it 😂

models/collection.js
r3rastogi: Ah interesting. So how would it look if we had multiple PRs open?

joelseq: @r3rastogi basically it will be dependent on the number of PR jobs. If there are multiple PRs open then there will be multiple PR jobs and the number of jobs would be represented by `prs.open` in the response. For each of the PR jobs, it will then look at the last build and the number of PRs whose last builds are failing would be represented by `prs.failing`.

PR #182, opened 2017-09-29
feat(723): allows requires keyword for jobs

d2lam:
Allows the keyword `requires`
Related: https://github.com/screwdriver-cd/screwdriver/issues/723

### DIFF
--- a/config/job.js
+++ b/config/job.js
@@ -65,9 +65,16 @@ const SCHEMA_SETTINGS = Joi.object().optional();
 const SCHEMA_STEP = Joi.alternatives().try(SCHEMA_STEP_STRING, SCHEMA_STEP_OBJECT);
 const SCHEMA_STEPS = Joi.array().items(SCHEMA_STEP).min(1);
 const SCHEMA_TEMPLATE = Joi.string().regex(Regex.FULL_TEMPLATE_NAME);
-
+const SCHEMA_JOBNAME = Joi.string().regex(Regex.JOB_NAME);
+const SCHEMA_TRIGGER = Joi.string().regex(Regex.TRIGGER);  // ~commit, ~pr, etc.
+const SCHEMA_REQUIRES_VALUE = Joi.alternatives().try(SCHEMA_JOBNAME, SCHEMA_TRIGGER);
+const SCHEMA_REQUIRES = Joi.alternatives().try(
+    Joi.array().items(SCHEMA_REQUIRES_VALUE),
+    SCHEMA_REQUIRES_VALUE
+);
 const SCHEMA_JOB = Joi.object()
     .keys({
+        requires: SCHEMA_REQUIRES,
         annotations: Annotations.annotations,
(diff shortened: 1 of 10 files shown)


config/job.js
```diff
@@ -65,9 +65,14 @@ const SCHEMA_SETTINGS = Joi.object().optional();
 const SCHEMA_STEP = Joi.alternatives().try(SCHEMA_STEP_STRING, SCHEMA_STEP_OBJECT);
 const SCHEMA_STEPS = Joi.array().items(SCHEMA_STEP).min(1);
 const SCHEMA_TEMPLATE = Joi.string().regex(Regex.FULL_TEMPLATE_NAME);
+const SCHEMA_REQUIRES = Joi.alternatives().try(
+    Joi.array().items(Joi.string()),
```
tkyi: Do we need to make sure it's `~commit`, `~pr`, or doesn't start with `~`? Or we don't care to validate the value in requires?

test/config/job.test.js
```diff
@@ -10,6 +10,10 @@ describe('config job', () => {
             assert.isNull(validate('config.job.job.yaml', config.job.job).error);
         });
 
+        it('validates a job with requires', () => {
+            assert.isNull(validate('config.job.job2.yaml', config.job.job).error);
```
tkyi: I'd prefer a more specific name than `job2`. Maybe `configv2`? Idk.

config/job.js
d2lam: I think it will be hard to verify it, since it could also be `requires: jobName`, and `jobName` can be any string

test/config/job.test.js
d2lam: kk 👍

config/job.js
d2lam: Correction: you're right. `jobName` is actually a regex, not just any string. So we can actually add some extra validation here.

test/config/job.test.js
```diff
@@ -10,6 +10,14 @@ describe('config job', () => {
             assert.isNull(validate('config.job.job.yaml', config.job.job).error);
         });
 
+        it('validates a job with requires', () => {
+            assert.isNull(validate('config.job.jobv2.yaml', config.job.job).error);
+        });
+
+        it('validates a job with requires', () => {
```
tkyi: does not validate*

test/data/validator.output.yaml
```diff
@@ -50,12 +50,10 @@ jobs:
...
   - image: node:4
+    requires: main
     commands:
     - name: install
       command: npm publish
     environment: {}
     secrets:
       - NPM_TOKEN
-
-workflow:
```
jithin1987: is there another test which makes use of `workflow` ?

d2lam: Let me separate the 2 tests instead of reusing

d2lam: @jithin1987  I added a separate test case

PR #183, opened 2017-10-06
feat: add optional startFrom field to event

minz1027:
- Add a optional `startFrom` field to event
- Add `create` event schema for new `POST /event` endpoint

Related: https://github.com/screwdriver-cd/screwdriver/issues/723

### DIFF
--- a/models/event.js
+++ b/models/event.js
@@ -4,6 +4,7 @@ const Joi = require('joi');
 const mutate = require('../lib/mutate');
 const Scm = require('../core/scm');
 const Workflow = require('../config/workflow');
+const { requiresValue } = require('../config/job');
 
 const MODEL = {
     id: Joi
@@ -32,6 +33,9 @@ const MODEL = {
         .string().hex().length(40)
         .description('SHA this project was built on')
         .example('ccc49349d3cffbd12ea9e3d41521480b4aa5de5f'),
+    startFrom: requiresValue
+        .description('Event start point - a job name or trigger name (~commit/~pr)')
+        .example('main'),
     type: Joi
         .string().valid([
             'pr',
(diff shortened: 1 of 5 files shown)


models/event.js
```diff
@@ -69,6 +73,18 @@ module.exports = {
...
+
+    /**
+     * Properties for Event that will be passed in a CREATE request
+     *
+     * @property create
+     * @type {Joi}
+     */
+    create: Joi.object(mutate(MODEL, [
+        'pipelineId', 'sha', 'type', 'workflow', 'startFrom'
+    ], [
```
d2lam: also need prRef

test/data/event.create.full.yaml
```diff
@@ -0,0 +1,9 @@
+# Event Create Example
+sha: 46f1a0bd5592a2f9244ca321b129902a06b53e03
```
jithin1987: Is `sha` a required value when creating event ? If it's empty will it default to HEAD ?

models/event.js
d2lam: nvm that would be pass to models

test/data/event.create.full.yaml
d2lam: I don't think it's needed from the API. We can look it up using getCommitSha

models/event.js
```diff
@@ -32,6 +33,9 @@ const MODEL = {
         .string().hex().length(40)
         .description('SHA this project was built on')
         .example('ccc49349d3cffbd12ea9e3d41521480b4aa5de5f'),
+    startFrom: requiresValue
```
jithin1987: Calling out job name will be problematic, now it's again up to the caller to infer. Can we call it `trigger` ? 
`trigger` can be `~commit` / `~pr`. This will just trigger all jobs which `requires` on `~commit` / `~pr`

Also `trigger` can be `job:jobName` (or something) if we want to support starting pipeline at a specific job.

d2lam: Not sure what you mean by "Calling out job name will be problematic"? 
requiresValue can take in both `jobName` or `~commit`

jithin1987: Scratch my comment, I misunderstood it as `startFrom` should be a `jobName`

models/event.js
```diff
@@ -32,6 +33,9 @@ const MODEL = {
...
+        .description('Job name the event starts from or ~commit/~pr')
```
jithin1987: Can you say something on the line `Event start point - a job name or trigger name (~commit/~pr)`

minz1027: addressed :)

models/event.js
petey: I'm not following the logic for why `trigger` is different from `startFrom` other than one has `job:name` syntax and the other does not.

PR #185, opened 2017-10-10
feat: add workflowGraph to pipeline and events

minz1027:
### Objective
Store graph generated by [workflow-parser](https://github.com/screwdriver-cd/workflow-parser) into a new optional field `workflowGraph` in both event and pipeline

- Add an optional field `workflowGraph` to `pipeline` and `event` model
- Allow `workflowGraph` in output generated by config-parser (used by API `validator`)

For legacy workflow config:
- Pipeline and events will have both workflow and workflowGraph

For new workflow config:
- Pipeline and events will only have workflowGraph

Related to https://github.com/screwdriver-cd/screwdriver/issues/723

### DIFF
--- a/config/workflowGraph.js
+++ b/config/workflowGraph.js
@@ -1,0 +1,22 @@
+'use strict';
+
+const Joi = require('joi');
+const { jobname, requiresValue } = require('../config/job');
+
+const SCHEMA_WORKFLOW_GRAPH = Joi.object().keys({
+    nodes: Joi.array().items(Joi.object().keys({
+        name: requiresValue
+    }).unknown()),
+    edges: Joi.array().items(Joi.object().keys({
+        src: requiresValue,
+        dest: jobname
+    }).unknown())
+});
+
+/**
+ * The definition of the Workflow Graph pieces
+ * @type {Object}
+ */
+module.exports = {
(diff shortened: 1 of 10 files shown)


config/workflowGraph.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const Joi = require('joi');
+const { jobname, requiresValue } = require('../config/job');
+
+const SCHEMA_WORKFLOW_GRAPH = Joi.object().keys({
+    nodes: Joi.array().items(Joi.object().keys({
+        name: requiresValue
```
petey: Node and edge objects may be too specific. Does this truncate keys that aren't listed? I'm certain we will want to add more data in here later.

minz1027: Yes, this will ignore keys that are not listed. At first, I was thinking if we want to add some other fields, we modify this data-schema to add it. But if we one this to be done once, I can make it just as an object.

tkyi: I agree with @minz1027. I like being very explicit with what we expect here so there's no magic or confusion.

minz1027: @petey Makes sense to allow us try out cool stuff without modifying data-schema every time. Changed it to allow unknown fields. https://github.com/hapijs/joi/blob/v11.3.4/API.md#objectunknownallow

PR #186, opened 2017-10-13
feat: allow PR-1:main as job name in model

minz1027:
- allow `PR-1:main` as job name in job model
   - in user yaml, jobNames start with `PR-` are not allowed so we do not take care of cases like `PR-1-main` in job model
- add a regex only for PR job name
  - will be used later to parse PR job name

### DIFF
--- a/config/regex.js
+++ b/config/regex.js
@@ -22,6 +22,8 @@ module.exports = {
     STEP_NAME: /^[\w-]+$/,
     // Jobs can only be named with A-Z,a-z,0-9,-,_
     JOB_NAME: /^[\w-]+$/,
+    // PR JOB Name can only be PR-1 or PR-1:main, group1: PR-prNum, group2: jobName
+    PR_JOB_NAME: /^(PR-\d+)(?::([\w-]+))?$/,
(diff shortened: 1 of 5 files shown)


config/regex.js
```diff
@@ -22,6 +22,8 @@ module.exports = {
     STEP_NAME: /^[\w-]+$/,
     // Jobs can only be named with A-Z,a-z,0-9,-,_
     JOB_NAME: /^[\w-]+$/,
+    // PR JOB Name can only be PR-1 or PR-1:main, gourp1: PR-prNum, group2: prNum, gourp3: jobName
```
tkyi: `group1`*

config/regex.js
```diff
@@ -22,6 +22,8 @@ module.exports = {
...
+    PR_JOB_NAME: /^(PR-(\d+))(?::([\w-]+))?$/,
```
tkyi: Is job name `meow-blah` still legal?
I don't think this regex captures that case

petey: Not sure why we need to parse out the pr number.

minz1027: Not really needed, taking it out.

PR #189, opened 2017-10-24
feat(770): allows sd@pipelineId:jobName for requires

d2lam:
We want to allow this so that the pipeline will be rebuilt when sd@pipelineId:jobName completes successfully

Related https://github.com/screwdriver-cd/screwdriver/issues/770

### DIFF
--- a/config/regex.js
+++ b/config/regex.js
@@ -24,8 +24,10 @@ module.exports = {
     JOB_NAME: /^[\w-]+$/,
     // PR JOB Name can only be PR-1 or PR-1:main, group1: PR-prNum, group2: jobName
     PR_JOB_NAME: /^(PR-\d+)(?::([\w-]+))?$/,
-    // Can be ~pr, ~commit, or ~commit:branchName
-    TRIGGER: /^~(pr|commit(:.+)?)$/,
+    // External trigger like ~sd@123:component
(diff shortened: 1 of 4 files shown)


config/regex.js
```diff
@@ -24,8 +24,8 @@ module.exports = {
     JOB_NAME: /^[\w-]+$/,
     // PR JOB Name can only be PR-1 or PR-1:main, group1: PR-prNum, group2: jobName
     PR_JOB_NAME: /^(PR-\d+)(?::([\w-]+))?$/,
-    // Can be ~pr, ~commit, or ~commit:branchName
-    TRIGGER: /^~(pr|commit(:.+)?)$/,
+    // Can be ~pr, ~commit, or ~commit:branchName, or ~sd@123:component
+    TRIGGER: /^~(sd@\d+:[\w-]+|pr|commit(:.+)?)$/,
```
minz1027: Can we also have a separate regex for  `~sd@\d+:[\w-]+` with capture groups? so that later we can use it to check if the trigger is external pipeline and get right groups we want out of it easily. e.g. pipelineId, jobName

d2lam: sure 👍

minz1027:
Tests for the regex under `regex.test.js`?

PR #190, opened 2017-10-25
feat(770): Add schema for trigger table

d2lam:
Add schema for the new trigger table.
No endpoint should be exposed to the users.

Related: https://github.com/screwdriver-cd/screwdriver/issues/770

### DIFF
--- a/models/trigger.js
+++ b/models/trigger.js
@@ -1,0 +1,60 @@
+'use strict';
+
+const Joi = require('joi');
+const Regex = require('../config/regex');
+
+const MODEL = {
+    id: Joi
+        .number().integer().positive()
+        .example(12345),
+
+    src: Joi
+        .string().regex(Regex.EXTERNAL_TRIGGER)
+        .example('~sd@1234:component'),
+
+    dest: Joi
+        .string().regex(Regex.EXTERNAL_TRIGGER)
+        .example('~sd@5678:test')
+
+};
+
+module.exports = {
+    /**
+     * All the available properties of Trigger
+     *
+     * @property base
+     * @type {Joi}
+     */
+    base: Joi.object(MODEL).label('Trigger'),
+
+    /**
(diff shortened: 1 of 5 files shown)


models/trigger.js
```diff
@@ -0,0 +1,62 @@
...
+        .example(12345),
+
+    pipelineId: Joi
+        .number().integer().positive()
+        .example(56789),
+
+    jobName: Job.jobname
+        .example('component'),
+
+    trigger: Joi
```
minz1027: Can we user the EXTERNAL_TRIGGER we just added to the regex? Just to be consistent

models/trigger.js
```diff
@@ -0,0 +1,63 @@
...
```
minz1027: `trigger` is a reserved keyword in a lot of SQL database. should probably user something else as the column name

d2lam: Changed the column name to `start` based on your suggestion :)

models/trigger.js
```diff
@@ -0,0 +1,60 @@
...
+const Regex = require('../config/regex');
+
+const MODEL = {
+    id: Joi
+        .number().integer().positive()
+        .example(12345),
+
+    src: Joi
+        .string().regex(Regex.EXTERNAL_TRIGGER)
+        .example('~sd@1234:component'),
```
jithin1987: why is `~ad@` syntax required here ? can't it be just `pipeline:jobName` format ?

tkyi: I think we're keeping `~sd@` for simplicity so we can reuse the `EXTERNAL_TRIGGER` regex to parse the `src` and `dest`.

minz1027: And I feel it's more consistent this way. Just in case we change the format we want for the EXTERNAL_TRIGGER later.

jithin1987: Sounds good.

PR #194, opened 2017-11-16
feat(694): add an optional statusMessage field to build model

minz1027:
add a optional statusMessage field to build model

Related to https://github.com/screwdriver-cd/screwdriver/issues/694

### DIFF
--- a/test/data/build.update.yaml
+++ b/test/data/build.update.yaml
@@ -1,4 +1,5 @@
 # Build Update Example
-status: ABORTED
+status: FAILURE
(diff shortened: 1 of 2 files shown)


test/data/build.update.yaml
```diff
@@ -1,4 +1,5 @@
 # Build Update Example
 status: ABORTED
+statusMessage: STOP
```
tkyi: Can you add a better example message? I'm still a little confused what it would look like.

PR #195, opened 2017-12-07
feat(api): Add command and commandTag data-schemas.

wdstar:
## Context

This PR adds `command` and `commandTag` data-schemas.
These schemas are based on the following `Commands` design document.

https://github.com/screwdriver-cd/screwdriver/blob/master/design/commands.md

## Goal (1st Phase: Minimum Viable Product)
- Commands REST API
  - post: create a command.
  - get: retrieve command details.
- **data-schemas** and data(factory)
  - elemental CRUD.

## Note

This PR blocks the implementations of the command data factory, the commands plugin (API endpoint) and the command validator.


models/command.js
```diff
@@ -0,0 +1,129 @@
...
+const mutate = require('../lib/mutate');
+const Command = require('../config/command');
+
+const MODEL = {
+    id: Joi
+        .number().integer().positive()
+        .description('Identifier of this command')
+        .example(123345),
+
+    labels: Joi
```
tk3fftk: What is `labels`? [design doc](https://github.com/screwdriver-cd/screwdriver/blob/master/design/commands.md) does not say about this.

wdstar: I add labels (tags) like the existing `template` schema.
https://github.com/screwdriver-cd/data-schema/blob/master/models/template.js
I guess that get api should return its command details with labels (tags).

test/data/config.command.bad.yaml
```diff
@@ -0,0 +1,15 @@
...
+version: '1.0'
+description: |
+  Lorem ipsum dolor sit amet.
+maintainer: foo@bar.com
+format: docker
+habitat:
+    mode: remote
+    package: core/git/2.14.1
+    binary: git
+#docker:
```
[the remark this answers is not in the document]
wdstar: It's not necessary.
But I'd like to test a format mismatch case that its format (`format: docker`) and command config (`habitat: ...`) are different.

test/data/config.commandFormat.binary.bad.yaml
```diff
@@ -0,0 +1,2 @@
+
```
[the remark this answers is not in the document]
wdstar: OK. Explicit null policy is a good habit. I will add an invalid line or a comment.

yoshwata:
Do you need to add explanations about new scheme on REAEDME?
https://github.com/screwdriver-cd/data-schema/blob/master/README.md

wdstar:
No. This readme seems to describe outline of the data-schema.
There is already a `models - Internal data resources (pipeline, job, build, collection etc.)` line.

yoshwata:
> No. This readme seems to describe outline of the data-schema.

I understand. Thank you.

test/data/config.command.bad.yaml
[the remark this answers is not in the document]
tk3fftk: I see. This case is looks good. I intended to say these lines seem to be not necessary in dummy data. Is it common?

test/data/command.update.yaml
```diff
@@ -0,0 +1,4 @@
+# Command UPDATE example
```
yoshwata: It might be better to comment that UPDATE or DELETE is not scope at this chance.

test/data/config.command.bad.yaml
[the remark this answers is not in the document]
wdstar: OK. I get it.
I deleted extra commented lines from test data and appended brief comments.

test/data/command.update.yaml
wdstar: I added this PR's goal (MVP) to my first comment.
Thanks.

yoshwata: I've checked and understood. Thank you.

models/command.js
```diff
@@ -0,0 +1,129 @@
...
+        .array()
+        .items(Joi.string())
+        .description('Labels for command')
+        .example(['stable', 'latest', 'beta']),
```
petey: Just being pedantic:
Better examples of labels would be 'docker', 'git', 'node'.

'stable', 'latest', 'beta' are tags.

wdstar: This examples follows the existing `template`'s for the avoidance of confusion.
https://github.com/screwdriver-cd/data-schema/blob/master/models/template.js#L18
Thanks.

petey: `template`'s `labels` is confusing, too. 😢

config/commandFormat.js
```diff
@@ -0,0 +1,66 @@
...
+    .requiredKeys('mode', 'package', 'binary');
+
+const DOCKER_IMAGE = Joi
+    .string()
+    .description('Image of the Docker command')
+    .example('node:1.2.3');
+
+const SCHEMA_DOCKER = Joi.object()
+    .keys({
+        image: DOCKER_IMAGE
```
catto: I think we should add `binary` as same as the schema of habitat, as an optional field and default "".
Some docker image have `entrypoint` or `command` set and others not.
When users want to run a SD command that uses latter one, they need to specify a command name explicitly.
> $ sd-cmd namespace/my-node node -v

If commands have binary field and a value has set, "node" in this case, users can use this command with the same parameter than other commands.
> $ sd-cmd namespace/my-node -v

wdstar: Spec. proposal:
https://github.com/screwdriver-cd/screwdriver/pull/826

wdstar:
Currently `command` object has `namespace` and `command` keys. redundant `command` !
Should we rname these keys?

Alternatives
1. single `name` key (like template): value examples; `'namespaece/command'` or `'command'`
1. better 2 key names:
    - `package`, `name`
    - `scope`, `name`
    - `group`, `name`
    - `namespace`, `name`
    - `namespace`, `commandName`
    - `namespace`, `commandname`
    - ...

models/command.js
d2lam: Yeah. I think we ended up using `tag` and `labels` is just there but not really used right now.

minz1027: Do not see a need to add `label` now. If we really want it, can do it in a second pass.

test/data/command.update.yaml
d2lam: We can probably remove this file

models/command.js
```diff
@@ -0,0 +1,129 @@
...
+        'labels'
+    ])).label('Create Command'),
+
+    /**
+     * Properties for command that will be passed during a UPDATE request
+     *
+     * @property update
+     * @type {Joi}
+     */
+    update: Joi.object(mutate(MODEL, [], ['labels']))
```
d2lam: We probably don't need this if we remove `labels`

config/command.js
```diff
@@ -0,0 +1,102 @@
...
+    .max(64)
+    .description('Maintainer of the Command')
+    .example('foo@bar.com');
+
+const COMMAND_FORMAT = Joi
+    .string().valid([
+        'habitat',
+        'docker',
+        'binary'
+    ])
```
tk3fftk: need `max()` for mysql?
https://github.com/screwdriver-cd/data-schema/pull/102

models/command.js
wdstar: Okay. I remove `labels` key at Phase 1.

config/commandFormat.js
wdstar: This proposal has been merged.
I'll rename and add these properties (habitat.command, docker.command).

models/command.js
wdstar: Yes, I'll remove it.

models/command.js
wdstar: done.

config/commandFormat.js
wdstar: done.

config/command.js
wdstar: I got it.  I added a length limitation to the  `command.format` property.
Thanks.

PR #196, opened 2017-12-25
fix: add missing API commandValidator schema.

wdstar:
## Context

This PR is a part of the `Screwdriver Commands` implementation.

## References

- [Design](https://github.com/screwdriver-cd/screwdriver/blob/master/design/commands.md)


yoshwata:
This pr is generated from private repository. Could you generate PR from sonic-screwdriver for someone in team to fix this pr.

wdstar:
I see.

api/commandValidator.js
```diff
@@ -0,0 +1,44 @@
...
+    })
+    .requiredKeys('message', 'path', 'type');
+
+const SCHEMA_OUTPUT = Joi.object()
+    .keys({
+        errors: Joi.array()
+            .items(COMMAND_ERROR)
+            .label('Array of errors encountered while validating the given command'),
+        // since a command could be parseable but invalid, the contents are unpredicatble
+        command: Joi.object()
```
tk3fftk: This should be `config` if we will implement `commandValidate` api like `templateValidate`.
https://github.com/screwdriver-cd/command-validator/blob/master/index.js#L54
When I implemented command-validator, I avoided using command because of key duplication.

We should fix response of command-validator to `command` from `config` if it's better.

wdstar: I guess your suggestion is better ...
But, if we use `config` instead of `command`, we will code `config.config` in `commands` plugin.
Ref.: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/templates/create.js#L39

We are in a dilemma.

d2lam: Yea, I think `command` is a little confusing since we have nested commands under it:
```
command:
    namespace: foo
    command: bar
    ...
    habitat:
        mode: remote
        package: core/git/2.14.1
        command: git
```

I feel like inside https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/templates/create.js#L27 you can just name that returned config something else, like `validConfig`. It's a variable in that block of code.

wdstar: `command` key rename PR has been merged.
https://github.com/screwdriver-cd/screwdriver/pull/829
This confusion will be resolved.
```
*command*:
    namespace: foo
    *name*: bar
    ...
    habitat:
        *command*: git
```
So, we can use `command` object instead of the existing `config` in the screwdriver-command-validator and use `config.command` as is in this API commandValidator schema.

d2lam: Ok got it 👍

PR #197, opened 2017-12-27
feat(827) support tolerations in executor start scheme

jithin1987:
# Context

To support passing toleration config for `executor-k8s-vm`

See https://github.com/screwdriver-cd/screwdriver/pull/828

### DIFF
--- a/plugins/executor.js
+++ b/plugins/executor.js
@@ -11,7 +11,8 @@ const SCHEMA_START = Joi.object().keys({
     apiUri: Joi.string().uri().required()
         .label('API URI'),
     token: Joi.string().required()
-        .label('Build JWT')
+        .label('Build JWT'),
+    tolerations: Joi.array().optional()
 }).required();
 const SCHEMA_STOP = Joi.object().keys({
     annotations: Annotations.annotations,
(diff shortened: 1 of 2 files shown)


plugins/executor.js
```diff
@@ -11,7 +11,8 @@ const SCHEMA_START = Joi.object().keys({
     apiUri: Joi.string().uri().required()
         .label('API URI'),
     token: Joi.string().required()
-        .label('Build JWT')
+        .label('Build JWT'),
+    tolerations: Joi.array().optional()
```
d2lam: We shouldn't need this change because of my comment here: https://github.com/screwdriver-cd/executor-k8s-vm/pull/19#pullrequestreview-85743527

jithin1987: I think this is required since, `executor-base` does schema validation.

d2lam: this schema is for this: https://github.com/screwdriver-cd/executor-k8s-vm/blob/master/index.js#L71-L80

We are not changing the method signature for `start`, so there is no need to change this schema.

jithin1987:
This change is not required.

PR #201, opened 2018-02-06
test: add match up test of indexes and rangeKeys.

wdstar:
## Context

Unit tests improvement.

## Objective

This PR adds match up test of `indexes` and `rangeKeys` properties about each model.

## Benefit

This test will prevent new added model's bug like this.
- https://github.com/screwdriver-cd/data-schema/pull/200

### DIFF
--- a/test/models/index.test.js
+++ b/test/models/index.test.js
@@ -5,13 +5,19 @@ const models = require('../../').models;
 
 describe('model commmons', () => {
     const modelsToCheck = [
-        'event',
-        'pipeline',
         'build',
-        'job',
-        'user',
+        'collection',
         'command',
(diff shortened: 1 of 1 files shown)


test/models/index.test.js
```diff
@@ -33,6 +39,19 @@ describe('model commmons', () => {
         });
     });
 
+    // See https://github.com/screwdriver-cd/datastore-sequelize/blob/master/index.js#L311
+    it('selected models have same lenght of indexes and rangeKeys.', () => {
```
tkyi: length*

wdstar: Oh, I fixed typo. Thanks.

PR #205, opened 2018-02-21
feat: allow status for creating build container

FenrirUnbound:
## Context

Before a build executes its first command, Screwdriver needs to perform a series of setup and configuration operations. Today, there isn't a specific state for describing these pre-build operations.

For this proposal, we aim to not solve the problem of describing every operation. We instead want to describe one state of all the setup operations.

## Objective

Enable a "creating container" build status in order to describe that a build is constructing the build container.

### DIFF
--- a/models/build.js
+++ b/models/build.js
@@ -106,6 +106,7 @@ const MODEL = {
 
     status: Joi
         .string().valid([
+            'CREATING_CONTAINER',
             'SUCCESS',
             'FAILURE',
             'QUEUED',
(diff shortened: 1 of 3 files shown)


models/build.js
```diff
@@ -106,6 +106,7 @@ const MODEL = {
 
     status: Joi
         .string().valid([
+            'CREATING_CONTAINER',
```
minz1027: Who is going to update the build status to `CREATING_CONTAINER `? The API set it to `QUEUED` and the build itself update it to `Running`. Who's going to be the middle man?

tkyi:
Not very useful to let the users know a container is creating vs QUEUED. Closing PR for now.

PR #208, opened 2018-03-07
fix(890): increase job name max length

tk3fftk:
Job name is too short and a shown error message "Data too long for column 'name' at row 1"  is hard to understand.

Related: https://github.com/screwdriver-cd/screwdriver/issues/890


tk3fftk:
I missed this comment
https://github.com/screwdriver-cd/screwdriver/issues/890#issuecomment-370999830
I'll increase max length to `128` from `50`. How about this?

PR #209, opened 2018-03-07
fix: Add event meta

tkyi:
Related to https://github.com/screwdriver-cd/screwdriver/issues/904

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ artifacts/
 npm-debug.log
 .DS_STORE
 .*.swp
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ artifacts/
 npm-debug.log
 .DS_STORE
 .*.swp
+package-lock.json
```
d2lam: I think Petey suggested before that we commit this file. 
We shouldn't ignore it?

PR #210, opened 2018-03-09
fix: Allow parentBuildId to be an array of ids

tkyi:
## Context
As a part of updating the metadata used in builds, we want to merge together meta from all parentBuildIds when running a join job.

## Objective
This PR allows the `parentBuildId` key to be a single number or an array of numbers (for backwards compatibility.
~This PR adds a `parentBuildIds` key which will be an array of parentBuildIds. Golang cannot handle multiple types for fields and we want to make sure our change is backwards compatible.~

Also adding `CREATED` status so the join build can be populated with `parentBuildId`s before starting.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/904

### DIFF
--- a/models/build.js
+++ b/models/build.js
@@ -110,11 +110,12 @@ const MODEL = {
 
     status: Joi
         .string().valid([
-            'SUCCESS',
-            'FAILURE',
-            'QUEUED',
             'ABORTED',
-            'RUNNING'
+            'CREATED', // when the build is created but not started
+            'FAILURE',
+            'QUEUED', // when the build is created and put into the queue
+            'RUNNING', // after the build is created, went through the queue, and has started
+            'SUCCESS'
         ])
         .description('Current status of the build')
         .example('SUCCESS')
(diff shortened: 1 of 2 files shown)


models/build.js
```diff
@@ -47,10 +48,13 @@ const MODEL = {
         .description('Identifier of the parent job')
         .example(123345),
 
-    parentBuildId: Joi
-        .number().integer().positive()
-        .description('Identifier of this parent build')
-        .example(123345),
+    parentBuildId: PARENT_BUILD_ID
+        .description('Identifier(s) of this parent build')
+        .example([123, 234]),
```
petey: This example should be just one numeric id.

d2lam:
Can we add another status value? maybe `CREATED`
The reason is, based on https://github.com/screwdriver-cd/screwdriver/pull/921, it will create a build without starting. If [A,B,C]->D. As soon as A is done, D will be created. And from the UI it will look like the build is QUEUED, which is really weird.

PR #212, opened 2018-03-19
fix: add an optional field `buildId` to event create schema

minz1027:
- add an optional field `buildId` to event create schema so that UI only need to send in `buildId` to restart a job
- will remove the stuff we don't need later, e.g. `parentBuildId`, `parentEventId` later. Just keep them here for now to make it backward compatible.


tkyi:
I think you're missing the test in this file:
https://github.com/screwdriver-cd/data-schema/blob/master/test/models/event.test.js

PR #213, opened 2018-03-21
fix(911): Add sourcePaths to job config and validator

tkyi:
## Context
Users don't always want builds to be triggered for all files in their repositories. Sometimes it would be nice if they could just configure certain source paths in their repos.

## Objective
This PR adds the `sourcePaths` keyword to job configs. It can either be an array of strings or a string.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/911

### DIFF
--- a/api/validator.js
+++ b/api/validator.js
@@ -27,9 +27,10 @@ const SCHEMA_JOB_PERMUTATION = Joi.object()
         description: Job.description,
         environment: Job.environment,
         image: Job.image,
+        requires: Job.requires,
         secrets: Job.secrets,
         settings: Job.settings,
-        requires: Job.requires
+        sourcePaths: Job.sourcePaths
     }).label('Job permutation');
 
 const SCHEMA_JOB_PERMUTATIONS = Joi.array().items(SCHEMA_JOB_PERMUTATION)
--- a/config/job.js
+++ b/config/job.js
@@ -72,16 +72,22 @@ const SCHEMA_REQUIRES = Joi.alternatives().try(
     Joi.array().items(SCHEMA_REQUIRES_VALUE),
     SCHEMA_REQUIRES_VALUE
 );
+const SCHEMA_SOURCEPATH = Joi.string().max(100).optional();
+const SCHEMA_SOURCEPATHS = Joi.alternatives().try(
+    Joi.array().items(SCHEMA_SOURCEPATH),
+    SCHEMA_SOURCEPATH
+);
 const SCHEMA_JOB = Joi.object()
     .keys({
-        requires: SCHEMA_REQUIRES,
         annotations: Annotations.annotations,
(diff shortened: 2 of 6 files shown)


api/validator.js
```diff
@@ -27,9 +27,10 @@ const SCHEMA_JOB_PERMUTATION = Joi.object()
         description: Job.description,
         environment: Job.environment,
         image: Job.image,
+        requires: Job.requires,
         secrets: Job.secrets,
         settings: Job.settings,
-        requires: Job.requires
+        sourcePaths: Job.sourcePaths
```
d2lam: Did we decide to have a separate keyword for this? or just use `settings`?

config/job.js
```diff
@@ -102,6 +108,8 @@ module.exports = {
     secret: SCHEMA_SECRET,
     secrets: SCHEMA_SECRETS,
     settings: SCHEMA_SETTINGS,
+    sourcePath: SCHEMA_SOURCEPATH,
+    sourcePaths: SCHEMA_SOURCEPATHS,
```
ScottyFillups: Why do we have `sourcePath` and `sourcePaths`? Is this for backward compatibility?

d2lam: It's a single `sourcePath` vs multiple `sourcePaths`. But in the yaml config, it only allows `sourcePaths`

Btw, can we change it to `srcpaths`? Then we don't need to get into the camel vs snake case

api/validator.js
jithin1987: separate keyword/job config, rather than in settings. #911 has details.

config/job.js
jithin1987: why do we have `sourcePath`, it should only be `sourcePaths`.

We don't need to consciously abbreviate to avoid camel case. As for camel vs snake case, if we have to use multi word keys stick with camel.

d2lam: It's just saying `sourcePaths` is an array of `sourcePath`, same as `secrets` is an array of `secret`.  
From the `screwdriver.yaml` point of view, the key is always `sourcePaths` (with an `s`) (line 90)

config/job.js
```diff
@@ -72,16 +72,22 @@ const SCHEMA_REQUIRES = Joi.alternatives().try(
     Joi.array().items(SCHEMA_REQUIRES_VALUE),
     SCHEMA_REQUIRES_VALUE
 );
+const SCHEMA_SOURCEPATH = Joi.string().max(100).optional();
+const SCHEMA_SOURCEPATHS = Joi.alternatives().try(
+    Joi.array().items(SCHEMA_SOURCEPATH),
+    SCHEMA_SOURCEPATH
```
petey: Are we really allowing sourcePaths to be either a string or an array? There's no test for this use case.

tkyi: Added a test

PR #214, opened 2018-03-22
fix: Add default value for meta field

tkyi:
## Context

There is a bug that causes builds to break if you run a detached job after an event occurs with no builds run (e.g., if someone makes a change to the screwdriver.yaml without triggering any builds). The code will try to marshal the meta and no meta will exist in the parent event.

## Objective

This PR adds a default empty object to `meta` in builds and events.

## Related links
- Related to the event meta work: https://github.com/screwdriver-cd/screwdriver/issues/904

### DIFF
--- a/models/build.js
+++ b/models/build.js
@@ -96,6 +96,7 @@ const MODEL = {
 
     meta: Joi
         .object()
+        .default({})
         .description('Key=>Value information from the build itself'),
 
     steps: Joi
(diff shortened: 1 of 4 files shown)


FenrirUnbound:
Is this testable?

tkyi:
We don't usually have tests for setting default values, but I've added some.

PR #215, opened 2018-03-24
feat(911): Add validation for scm methods parseHook and getChangedFiles [1]

tkyi:
## Context
We'd like to add a method to scm-base that gets a list of changed files when a push or PR occurs in Git. This will allow for subdirectory support.

## Objective
This PR:
- alphabetizes the keys in the **parseHook** schema
- adds `lastCommitMessage` and `scmContext` to the parseHook schema
- adds input/output validation for **parseHook** and **getFilesChanged** methods

## Related links
- Related to https://github.com/screwdriver-cd/screwdriver/issues/911

### DIFF
--- a/core/scm.js
+++ b/core/scm.js
@@ -74,15 +74,6 @@ const SCHEMA_COMMIT = Joi.object().keys({
 }).label('SCM Commit');
 
 const SCHEMA_HOOK = Joi.object().keys({
-    type: Joi.string()
-        .valid(['pr', 'repo', 'ping'])
-        .required()
-        .label('Type of the event'),
-
-    hookId: Joi.string()
-        .required()
-        .label('Uuid of the event'),
-
     action: Joi.string()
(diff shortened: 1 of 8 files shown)


core/scm.js
```diff
@@ -115,15 +118,22 @@ const SCHEMA_HOOK = Joi.object().keys({
         .optional()
         .label('PR original source'),
 
-    prRef: Joi.string()
-        .optional()
-        .label('PR reference of the repository'),
+    scmContext: Joi
+        .string().max(128)
+        .required()
```
FenrirUnbound: If this is required, can this change still be backwards-compatible?

tkyi: We weren't using any validation before for parseHook, so the entire `hook` schema was never used a whole, just parts of it.
I was thinking of just bumping major for scm-base and scm-github after we pull in/use the new validation, but this change (in the data-schema) shouldn't break anything.

PR #220, opened 2018-04-02
fix: Allow null for hookId

tkyi:
## Context
scm-gitlab returns null for hookId.

## Objective
This PR allows null for hookId.

## Related links
scm-gitlab: https://github.com/tkyi/scm-gitlab/blob/master/index.js#L320
Original issue: https://github.com/screwdriver-cd/screwdriver/issues/911


tkyi:
It looks like `hookId` is just used for logging (https://github.com/screwdriver-cd/screwdriver/blob/094339fd49a0b266438fe3adf7d22e3675eafba5/plugins/webhooks/index.js#L102). Changing it to optional.

PR #221, opened 2018-04-04
 fix(hip): Update testing environment to node:8

ScottyFillups:
## Context

The Screwdriver API repository uses `node:8` as its image; ideally the same image should be used in this repository's `screwdriver.yaml`. 

In addition, this repository uses an outdated version of `eslint-config-screwdriver`, and has different ESLint rules from the other Screwdriver repositories.

## Objective

* Bump the Node.js version in `screwdriver.yaml` to `node:8`
* Update ESLint, and lint the files with the new rules


tkyi:
You'll need to be careful because everything that depends on this package will probably need to be updated to node8.

screwdriver.yaml
```diff
@@ -1,19 +1,15 @@
...
         steps:
             - install: npm install
             - test: npm test
-
     publish:
-        steps:
-            - install: npm install --no-save semantic-release@^7
-            - publish: npm run semantic-release
+        requires: main
+        template: screwdriver-cd/semantic-release-node8@1.0.0
```
minz1027: this template doesn't work and we haven't got time to look into this, you can still use the old template https://github.com/screwdriver-cd/models/blob/master/screwdriver.yaml#L13 to publish

ScottyFillups: Done! 😄

PR #226, opened 2018-04-17
fix: Fix Joi.reach call to work with new Joi version

pranavrc:
# Context

We pin our Joi version to `13.1.2` in `package-lock.json` but the latest release of joi breaks one of our lines of code because it changes the behavior of `Joi.reach` to accept arrays.

# Objective

- [x] Fix the broken `joi.reach` call to use an array separator instead of string.

# Related links

Original issue in the joi repo: https://github.com/hapijs/joi/issues/1383
PR that changes the behavior of joi.reach: https://github.com/hapijs/joi/pull/1433


pranavrc:
Note that this change won't work with versions of Joi previous to `13.2.0` though, so if we're rolling this out, we need to roll it out with the new Joi version `13.2.0`.

pranavrc:
Never mind, changing the line to not use Joi at all, since using Joi.reach with an empty string seems kinda pointless.

PR #227, opened 2018-04-19
Banners

russdeluca:


models/banner.js
```diff
@@ -0,0 +1,68 @@
...
+const Joi = require('joi');
+const mutate = require('../lib/mutate');
+
+const MODEL = {
+    id: Joi
+        .number()
+        .integer()
+        .positive(),
+
+    message: Joi
```
minz1027: Can we add a field to record the time when this created? like this https://github.com/screwdriver-cd/data-schema/blob/913538c6de39a0c6edb2fa1f2068e1766d81f711/models/build.js#L78

tkyi:
Can you add a description with a link to the issue?

models/banner.js
d2lam: That will be helpful

models/banner.js
```diff
@@ -0,0 +1,68 @@
...
+     */
+    base: Joi.object(MODEL).label('Banner'),
+
+    /**
+     * Properties for User that will be passed during a CREATE request
+     *
+     * @property create
+     * @type {Joi}
+     */
+    create: Joi.object(mutate(MODEL, [
```
d2lam: I imagine you will have an `update` too right? (for people to update old banner)
And maybe `get or `list`, depending on what you'll need it on the API side.

tkyi: I think Min and I were thinking just an update (`PUT`) and delete for the first pass. https://github.com/screwdriver-cd/screwdriver/issues/966#issuecomment-378743186

minz1027: I think we need get and list for the UI to render it

d2lam:
Also, commit message will need to be in this form since we use `semantic-release`: https://github.com/screwdriver-cd/data-schema/blob/master/CONTRIBUTING.md#commit-message-format

models/banner.js
```diff
@@ -0,0 +1,130 @@
...
+    ])).label('Create Banner'),
+
+    /**
+     * Properties for Banners that will be passed during a UPDATE request
+     *
+     * @property update
+     * @type {Joi}
+     */
+    update: Joi.object(mutate(MODEL, [], [
+        'id',
```
tkyi: probably don't want let users update `id`

models/banner.js
```diff
@@ -0,0 +1,130 @@
...
+    base: Joi.object(MODEL).label('Banner'),
+
+    /**
+     * Properties for Banner that will come back during a GET request
+     *
+     * @property get
+     * @type {Joi}
+     */
+    get: Joi.object(mutate(MODEL, [
+        'id'
```
tkyi: If these are fields that are returned during a GET request, shouldn't it at least also include everything passed in for the CREATE (`message`, `type`)?

models/banner.js
```diff
@@ -0,0 +1,130 @@
...
+        'type',
+        'isActive'
+    ])).label('Update Banner'),
+
+    /**
+     * Properties for Banners that will come back during a LIST request.
+     * The LIST request will list all banners
+     */
+    list: Joi.array().items(Joi.object(mutate(MODEL, [
+    ], [
```
tkyi: should probably at least return `id`
should be similar to https://github.com/screwdriver-cd/data-schema/blob/9f9d6fbf23716b278e7b0ad7ed3095973c7b2bda/models/collection.js#L76-L83

tkyi: Can also remove this section in here and just do something like https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/templates/list.js#L6 in the API repo later

models/banner.js
```diff
@@ -0,0 +1,130 @@
...
+        .description('Body of banner to display')
+        .example('Due to planned upgrade of Kubernetes, Screwdriver will be down'),
+
+    isActive: Joi
+        .boolean()
+        .description('Flag if the banner is active')
+        .example(true)
+        .default(false),
+
+    dateCreated: Joi
```
tkyi: Just for consistency, I think `createTime` is what we've been using

models/banner.js
```diff
@@ -0,0 +1,130 @@
...
+        .positive()
+        .description('ID of user creating the banner'),
+
+    type: Joi
+        .string().valid([
+            'info',
+            'warn'
+        ])
+        .description('Type/Severity of the banner message')
+        .example('info')
```
tkyi: can set a default value if you want it to be optional

models/banner.js
```diff
@@ -0,0 +1,130 @@
...
+        .string()
+        .isoDate()
+        .description('When this banner was created')
+        .example('2017-01-06T01:49:50.384359267Z'),
+
+    createdBy: Joi
+        .number()
+        .integer()
+        .positive()
+        .description('ID of user creating the banner'),
```
tkyi: I don't think we have user ID numbers. It'll have to be username https://github.com/screwdriver-cd/data-schema/blob/be07dfdab5d6e4ede1c46a8bec5dd213ec512e97/models/user.js#L12-L16

models/banner.js
```diff
@@ -0,0 +1,139 @@
...
+        .string()
+        .max(512)
+        .description('Body of banner to display')
+        .example('Due to planned upgrade of Kubernetes, Screwdriver will be down'),
+
+    isActive: Joi
+        .boolean()
+        .description('Flag if the banner is active')
+        .example(true)
+        .default(false),
```
minz1027: Just as a reminder, setting default value here is not going to automatically generate the value for this field in DB. If we really want this to be default value in the DB, we still need to set it in model create part. This will generate the value only when you do joi.validate with the schema, but we're not using the joi object coming from joi.validate when we save it...

PR #228, opened 2018-04-25
feat(periodicBuilds): Schema for executor.startPeriodic and executor.stopPeriodic

pranavrc:
# Context

To implement functions for periodic starts and stops on the executor-base, we need schemas to provide validation.

# Objective

This PR implements executor.startPeriodic and executor.stopPeriodic schemas that can be used in the corresponding executor-base functions.

# Related links

executor-base PR: https://github.com/screwdriver-cd/executor-base/pull/39
executor-queue PR: https://github.com/screwdriver-cd/executor-queue/pull/12
Issue that tracks the periodic build feature: https://github.com/screwdriver-cd/screwdriver/issues/688

### DIFF
--- a/plugins/executor.js
+++ b/plugins/executor.js
@@ -17,6 +17,15 @@ const SCHEMA_STOP = Joi.object().keys({
     annotations: Annotations.annotations,
     buildId
 }).required();
+const SCHEMA_START_PERIODIC = Joi.object().keys({
+    pipeline: Joi.object().required(),
+    job: Joi.object().required(),
+    tokenGen: Joi.func().required(),
+    update: Joi.boolean().required()
+}).required();
+const SCHEMA_STOP_PERIODIC = Joi.object().keys({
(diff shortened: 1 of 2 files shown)


plugins/executor.js
```diff
@@ -17,6 +17,15 @@ const SCHEMA_STOP = Joi.object().keys({
     annotations: Annotations.annotations,
     buildId
 }).required();
+const SCHEMA_START_PERIODIC = Joi.object().keys({
+    pipeline: Joi.object().required(),
```
tkyi: Should `pipeline` and `job` use schemas from the models? (https://github.com/screwdriver-cd/data-schema/blob/master/models/pipeline.js#L65, https://github.com/screwdriver-cd/data-schema/blob/master/models/job.js#L54)

d2lam:
I think we decided to not add this. Since the schema changes based on how other executors implement their periodically builds. For example, Jenkins might not need all this input. So we won't do validation anymore. 
Can we close this PR?

PR #230, opened 2018-05-04
make Joi for PR key in event object

lusol:
adding verfication for PR field in event object

Related to https://github.com/screwdriver-cd/screwdriver/issues/822

### DIFF
--- a/core/scm.js
+++ b/core/scm.js
@@ -76,7 +76,6 @@ const SCHEMA_COMMIT = Joi.object().keys({
 const SCHEMA_PR = Joi.object().keys({
     url: Joi.string()
         .uri()
-        .required()
         .label('Link to PR')
         .example('https://git.ouroath.com/MAILSERVICES/gdpr-event-processor/pull/1')
 }).label('SCM Pull Request');
--- a/core/scm.js
+++ b/core/scm.js
@@ -73,6 +73,14 @@ const SCHEMA_COMMIT = Joi.object().keys({
         .example('https://github.com/scredriver-cd/screwdriver/commit/8843d7f92416211de')
 }).label('SCM Commit');
 
+const SCHEMA_PR = Joi.object().keys({
+    url: Joi.string()
+        .uri()
+        .required()
+        .label('Link to PR')
+        .example('https://git.ouroath.com/MAILSERVICES/gdpr-event-processor/pull/1')
+}).label('SCM Pull Request');
+
 const SCHEMA_HOOK = Joi.object().keys({
     action: Joi.string()
         .when('type', { is: 'pr',
@@ -149,5 +157,6 @@ module.exports = {
     commit: SCHEMA_COMMIT,
     repo: SCHEMA_REPO,
     user: SCHEMA_USER,
-    hook: SCHEMA_HOOK
+    hook: SCHEMA_HOOK,
+    pr: SCHEMA_PR
 };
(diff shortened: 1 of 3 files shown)


models/event.js
```diff
@@ -60,7 +60,11 @@ const MODEL = {
         .example({
             nodes: [{ name: '~commit' }, { name: 'main' }, { name: 'publish' }],
             edges: [{ src: '~commit', dest: 'main' }, { src: 'main', dest: 'publish' }]
-        })
+        }),
+    pr: Joi
```
minz1027: Need to add this as an optional field in the get schema, so that the API will allow this field to be returned during a GET event request

minz1027: And probably better to make  `pr` an object and have `url` as a field inside. Later on if we want to add new stuff related to pr to the event, then we can easily do that by adding stuff to this object.

models/event.js
```diff
@@ -60,7 +60,12 @@ const MODEL = {
         .example({
             nodes: [{ name: '~commit' }, { name: 'main' }, { name: 'publish' }],
             edges: [{ src: '~commit', dest: 'main' }, { src: 'main', dest: 'publish' }]
+        }),
+    pr: Joi
+        .object().keys({
+            url: Joi.string().description('url for PR link').example('https://git.ouroath.com/MAILSERVICES/gdpr-event-processor/pull/1')
```
tkyi: Can you put `.description` and `.example` on newlines so it's a little easier to read?

tkyi:
You have an eslint error: 
```
11:13:03
 /sd/workspace/src/github.com/screwdriver-cd/data-schema/models/event.js
11:13:03
   66:1  error  Line 66 exceeds the maximum line length of 100  max-len
```

models/event.js
minz1027: can use Joi.string().uri() instead, otherwise lgtm 👍

models/event.js
```diff
@@ -60,7 +60,16 @@ const MODEL = {
         .example({
             nodes: [{ name: '~commit' }, { name: 'main' }, { name: 'publish' }],
             edges: [{ src: '~commit', dest: 'main' }, { src: 'main', dest: 'publish' }]
+        }),
+    pr: Joi
```
tkyi: Since you're putting the same Joi schema in the `scm` already, you can just put `pr: Scm.pr` here (like https://github.com/screwdriver-cd/data-schema/blob/master/models/event.js#L25) instead of duplicating the schema.

plugins/scm.js
```diff
@@ -107,6 +108,10 @@ const DECORATE_AUTHOR = Joi.object().keys({
     scmContext
 }).required();
 
+const DECORATE_PR = Joi.object().keys({
+    prURL
```
tkyi: I don't think you want to take in a PR URL for the `decoratePR` method.

plugins/scm.js
```diff
@@ -107,6 +107,13 @@ const DECORATE_AUTHOR = Joi.object().keys({
     scmContext
 }).required();
 
+const DECORATE_PR = Joi.object().keys({
```
tkyi: don't need anything in this file anymore

core/scm.js
```diff
@@ -73,6 +73,14 @@ const SCHEMA_COMMIT = Joi.object().keys({
         .example('https://github.com/scredriver-cd/screwdriver/commit/8843d7f92416211de')
 }).label('SCM Commit');
 
+const SCHEMA_PR = Joi.object().keys({
+    url: Joi.string()
+        .uri()
+        .required()
```
tkyi: Can get rid of `required()` here

PR #239, opened 2018-05-25
[1] feat(937): Add config pipeline fields and scmUrl schema

Filbird:
**Update 5/30**

Add `configPipelineSha` field to the events model. 

---
**Update 5/29**

Replace `isConfigPipeline` field with `scmUrls`.

---

Add optional fields `configPipelineId` and `isConfigPipeline` to the pipeline model to distinguish between **config**, **child**, and **regular** pipelines (unsure how to officially name each of these types of pipelines).

`configPipelineId` is the id of the pipeline that contains the configuration for this pipeline. This field will only be populated for child pipelines.

`isConfigPipeline` is true if the pipeline is a config pipeline.

We should discuss if adding these fields is the correct approach to distinguish between different pipeline types.

---

Also add `scmUrls ` schema into the pipeline level config.

https://github.com/screwdriver-cd/screwdriver/issues/937

### DIFF
--- a/models/event.js
+++ b/models/event.js
@@ -44,6 +44,10 @@ const MODEL = {
         .string().hex().length(40)
         .description('SHA this project was built on')
         .example('ccc49349d3cffbd12ea9e3d41521480b4aa5de5f'),
+    configPipelineSha: Joi
+        .string().hex().length(40)
+        .description('SHA of the configuration pipeline this project depends on')
+        .example('ccc49349d3cffbd12ea9e3d41521480b4aa5de5f'),
     startFrom: Joi
         .string()
         .description('Event start point - a job name or trigger name (~commit/~pr)')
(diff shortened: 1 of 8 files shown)


models/pipeline.js
```diff
@@ -52,7 +52,15 @@ const MODEL = {
...
     lastEventId: Joi.number().integer().positive()
         .description('Identifier of last event')
-        .example(123345)
+        .example(123345),
+
+    configPipelineId: Joi.number().integer().positive()
+        .description('Identifier of pipeline containing external configuration')
+        .example(123),
+
+    isConfigPipeline: Joi.boolean()
```
jithin1987: can use use presence of `scmUrls` to infer this ?

Filbird: yep, makes sense!

minz1027: Then need to add `scmUrls` to this pipeline model @

models/event.js
```diff
@@ -44,6 +44,10 @@ const MODEL = {
         .string().hex().length(40)
         .description('SHA this project was built on')
         .example('ccc49349d3cffbd12ea9e3d41521480b4aa5de5f'),
+    configPipelineSha: Joi
+        .string().hex().length(40)
+        .description('SHA of the configurgation pipeline this project depends on')
```
tkyi: configuration*

PR #244, opened 2018-05-30
fix: Remove Joi alternatives from template model

tkyi:
Models can't handle Joi alternatives. Hardcoding template name to a string instead.

Seeing this error when trying to run the API locally:
```
TypeError: Cannot read property 'key' of undefined
```

Related to https://github.com/screwdriver-cd/screwdriver/issues/926#issuecomment-391907638

### DIFF
--- a/models/template.js
+++ b/models/template.js
@@ -19,7 +19,11 @@ const MODEL = {
 
     config: Template.config,
     namespace: Template.namespace,
-    name: Template.name,
+    name: Joi
+        .string()
+        .max(64)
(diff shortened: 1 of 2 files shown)


models/template.js
```diff
@@ -19,7 +19,11 @@ const MODEL = {
 
     config: Template.config,
     namespace: Template.namespace,
-    name: Template.name,
```
jithin1987: should we remove `Template.name` from `config/`

PR #247, opened 2018-06-01
fix(1082): [6] add regex validation of commit branch filter.

wdstar:
## Context

SCM Branch-specific jobs

## Objective

This PR adds  regex validation of commit branch filter (e.g. `/^user-.*$/`).

## References

- [Design PR](https://github.com/screwdriver-cd/screwdriver/pull/1092)
- https://github.com/screwdriver-cd/screwdriver/issues/1082

### DIFF
--- a/config/job.js
+++ b/config/job.js
@@ -4,6 +4,41 @@ const Annotations = require('./annotations');
 const Joi = require('joi');
 const Regex = require('./regex');
 
+// ref. https://github.com/hapijs/joi/blob/v13.3.0/API.md#extendextension
+const sdJoi = Joi.extend(joi => ({
+    base: joi.string(),
+    name: 'string',
+    language: {
+        commitBranch: 'needs to be a commit branch string/regex'
+    },
+    rules: [
+        {
+            name: 'commitBranch',
+            validate(params, value, state, options) {
+                const matched = Regex.TRIGGER.exec(value);
+                const brFilter = matched[3];
+
+                // branch regex filter
+                if (typeof brFilter !== 'undefined' && /^\/.+\/$/.test(brFilter)) {
+                    try {
+                        const filterRegex = brFilter.substr(1, brFilter.length - 2);
+                        /* eslint-disable no-unused-vars */
+                        // compile for syntax validation
+                        const re = new RegExp(filterRegex);
+                        /* eslint-enable */
+                    } catch (e) {
+                        return this.createError(
+                            'string.commitBranch', { v: value, err: e.message }, state, options);
+                    }
+                }
+
+                // no filter or exact match string
+                return value;
+            }
+        }
+    ]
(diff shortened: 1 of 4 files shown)


config/regex.js
```diff
@@ -49,6 +49,8 @@ module.exports = {
     // Can be ~pr, ~commit, or ~commit:branchName, or ~sd@123:component
     // Note: if you modify this regex, you must modify `sdJoi` definition in the `config/job.js`
     TRIGGER: /^~(sd@\d+:[\w-]+|pr|commit(:(.+))?)$/,
+    // Can be ~commit or ~commit:master
```
yokawara: `~commit:branchName` is better.

yuichi10: Thanks I will fix!

config/job.js
```diff
@@ -4,6 +4,41 @@ const Annotations = require('./annotations');
...
+    name: 'string',
+    language: {
+        commitBranch: 'needs to be a commit branch string/regex'
+    },
+    rules: [
+        {
+            name: 'commitBranch',
+            validate(params, value, state, options) {
+                const matched = Regex.TRIGGER.exec(value);
+                const brFilter = matched[3];
```
tk3fftk: It seems to be a magic number.

config/job.js
```diff
@@ -4,6 +4,42 @@ const Annotations = require('./annotations');
...
+    language: {
+        commitBranch: 'needs to be a commit branch string/regex'
+    },
+    rules: [
+        {
+            name: 'commitBranch',
+            validate(params, value, state, options) {
+                const regexPos = 3;
+                const matched = Regex.TRIGGER.exec(value);
+                const brFilter = matched[regexPos];
```
tk3fftk: I think we should also add comments to these parsing.

yuichi10: Sorry my variable name is not good I will fix the name regexPos to specificBranchPos.
Then, it give you sense which part this code try to get. 
After this correction and see the Regex.TRIGGER, maybe you can understand easily.

What do you think?

yuichi10: I added comments.

config/job.js
```diff
@@ -4,6 +4,44 @@ const Annotations = require('./annotations');
...
+    base: joi.string(),
+    name: 'string',
+    language: {
+        commitBranch: 'needs to be a commit branch string/regex'
+    },
+    rules: [
+        {
+            name: 'commitBranch',
+            validate(params, value, state, options) {
+                const specificBranchPos = 3;
```
tkyi: I think it's more clear this is a constant, hardcoded value if you put it in all caps like https://github.com/screwdriver-cd/executor-k8s-vm/blob/master/index.js#L14

yuichi10: I fixed!

PR #250, opened 2018-06-06
feat(614): [1] add pipelineId to token model

s-yoshika:
## Context
To implement pipeline API token.

## Objective
I added `pipelineId` to token model.
`pipelineId` is used to indicate which pipeline the token belongs to.
`userId` and `pipelineId` is alternative keys.

token model is currently used as user token, and the token model has a `userId` to indicate which user it belongs to, but this is not required. So we can determine the role of the token depending on whether token has `userId` or `pipelineId`. It doesn't break compatibility with currently used user tokens.

## References
proposal: https://github.com/screwdriver-cd/screwdriver/issues/614

### DIFF
--- a/models/token.js
+++ b/models/token.js
@@ -27,6 +27,13 @@ const MODEL = {
         .positive()
         .description('User ID'),
 
+    pipelineId: Joi
+        .number()
+        .integer()
+        .positive()
+        .description('Pipeline ID')
+        .example(123345),
+
     name: Joi
         .string()
         .max(128)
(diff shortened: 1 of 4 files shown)


models/token.js
```diff
@@ -54,7 +61,7 @@ module.exports = {
      * @property base
      * @type {Joi}
      */
-    base: Joi.object(MODEL).label('Token'),
+    base: Joi.object(MODEL).without('userId', 'pipelineId').label('Token'),
```
tk3fftk: It passed when both `userid` and `pipelienId` are absent. Is this ok? If not, how about using [xor] (https://github.com/hapijs/joi/blob/7a0582d1b7b473b5d7e3f7f4e97beaf918f1229c/lib/types/object/index.js#L484).

s-yoshika: Originally, `userId` is not required, so I can't decide it should be required.
https://github.com/sonic-screwdriver-cd/data-schema/blob/5979c9237bcf6c8fdf56418576c3d30e05f7f9eb/models/token.js#L24

tk3fftk: 🙆

PR #251, opened 2018-06-06
fix(937): Add startAll flag and scmUrls to childPipelines object

Filbird:
## Context
Beyond just listing `scmUrls` we want to enable additional configuration within the `screwdriver.yaml` for  config pipelines. For instance, we want to include a `startAll` flag to determine whether to start all child pipelines whenever  the config pipeline starts.

## Objective
Encompass `scmUrls` and `startAll` into an object named `childPipelines`.

## Reference
https://github.com/screwdriver-cd/screwdriver/issues/937

### DIFF
--- a/api/validator.js
+++ b/api/validator.js
@@ -49,7 +49,7 @@ const SCHEMA_OUTPUT = Joi.object()
         annotations: Annotations.annotations,
         errors: Joi.array().items(Joi.string()).optional(),
         jobs: SCHEMA_JOBS,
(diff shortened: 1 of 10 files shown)


Filbird:
Hrmm. I agree, it's a little ambiguous. I do like `childPipelines` as an alternative.

Thoughts @minz1027  @jithin1987 ?

tkyi:
Wondering if `configPipelineId` would ever be set when `childPipelines` exists. Should we make sure they can't coexist?

PR #260, opened 2018-06-26
feat(1132): Add parent config schema to getCheckoutCommand

Filbird:
## Context
In order for an `scm` to clone the parent repository of a child pipeline, it must have information on the parent pipeline. 

This configuration will be used by `getSetupCommand` of `scm-base`.

## Objective
Include `PARENT_CONFIG` schema in `GET_CHECKOUT_COMMAND`.

## Reference
https://github.com/screwdriver-cd/screwdriver/issues/1132

### DIFF
--- a/plugins/scm.js
+++ b/plugins/scm.js
@@ -29,6 +29,12 @@ const ADD_WEBHOOK = Joi.object().keys({
     scmContext
 }).required();
 
+const PARENT_CONFIG = Joi.object().keys({
+    branch: Joi.string().required(),
+    host: Joi.string().required(),
+    org: Joi.string().required(),
+    repo: Joi.string().required()
+});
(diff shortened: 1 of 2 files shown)


plugins/scm.js
```diff
@@ -29,6 +29,12 @@ const ADD_WEBHOOK = Joi.object().keys({
     scmContext
 }).required();
 
+const PARENT_CONFIG = Joi.object().keys({
+    branch: Joi.string().required(),
+    host: Joi.string().required(),
+    org: Joi.string().required(),
+    repo: Joi.string().required()
```
minz1027: You probably also need `sha` here for restarting past build from the child pipeline.

Filbird: Good point! Will add

PR #261, opened 2018-07-09
fix(1095): Add command arguments field [1]

tkyi:
## Context
It would be nice if users could add optional arguments to their command usage.

## Objective
This PR adds an optional field "arguments" to commands.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1095


config/command.js
```diff
@@ -45,6 +45,13 @@ const COMMAND_DESCRIPTION = Joi
     .description('Description of the Command')
     .example('Installs npm modules');
 
+const COMMAND_ARGUMENTS = Joi
+    .string()
+    .max(256)
```
tk3fftk: Is 256 enough?  I imagine the arguments view can be like CLI command.
```
Usage: sd-cmd exec foo/bar@1.2.3 <OPTION> <TARGET>
Options:
        --config               some config file
        --debug               debug mode (default "false")
        --host                  super host
        --log-level           Set the logging level ("debug"|"info"|"warn"|"error"|"fatal") (default "info")
Target:
...
```

test/data/command.create.yaml
```diff
@@ -4,7 +4,14 @@ name: bar
 version: '1.0'
 description: |
   Lorem ipsum dolor sit amet.
-arguments: '-d <domain> -h <host>'
+arguments: '<OPTION> <TARGET>
```
tk3fftk: Do we need `|` to keep new lines?

tkyi: Newlines are not respected by the UI, so I always just assumed we used it for easier readability in the yaml file.

tkyi:
@d2lam `usage`? But I'm worried people will think they need to write the whole `sd-cmd exec foo/bar@1.2.3`

tkyi:
**Update 7/16/2018:** 
Changed from `arguments` to `usage`. If `usage` field is set, it will override the usage field in the UI. Otherwise, the default will be used (as before).

test/data/command.create.yaml
tkyi: Sorry I just realized you meant I should add `|`
Yes, you need to add `|` for new lines.

PR #265, opened 2018-07-20
fix(1200): Fix git checkout url parsing regex [1]

tkyi:
## Context
The checkout url allows invalid git urls to be passed in, like `git remote add origin git@github.com:elyzion/myRepo.git`.

## Objective
This PR updates the regex for checkoutUrl parsing by disallowing spaces.

## Relevant links
Issue: https://github.com/screwdriver-cd/screwdriver/issues/1200


config/regex.js
```diff
@@ -59,7 +59,7 @@ module.exports = {
     ENV_NAME: /^[A-Z_][A-Z0-9_]*$/,
     // Repo checkout url. For example: https://github.com/screwdriver-cd/data-schema.git#branchName or git@github.com:screwdriver-cd/data-schema.git
     // eslint-disable-next-line max-len
-    CHECKOUT_URL: /^(?:(?:https?|git):\/\/)?(?:[^@]+@)?([^/:]+)(?:\/|:)([^/]+)\/(.+?)(?:\.git)?(#.+)?$/,
+    CHECKOUT_URL: /^(?:(?:https:\/\/(?:[^@/:\s]+@)?)|git@)+([^/:\s]+)(?:\/|:)([^/:\s]+)\/([^\s]+?)(?:\.git)?(#[^\s]+)?$/,
```
Filbird: Seems like this regex captures input such as, `https://fjahja@github.com:org/repo.git` and treats `https://fjahja@` as a single group.

I think we can simplify the first part of the regex from `(?:(?:https:\/\/(?:[^@/:\s]+@)?)|git@)+` to 
`(?:https:\/\/|[^@/:\s]+@)`.

tkyi: That regex won't work for bitbucket checkoutUrls under a team/org like `https://tkyi@bitbucket.org/d2lam2/test.git#branch`

PR #268, opened 2018-08-03
fix(1114): Add step model

tkyi:
## Context
It's hard to prevent race conditions when Step is a single field in the Build model. When trying to update the `endTime` and `lines` fields at the same time from the launcher and the log-service, the entire Step field can be overwritten, causing a loss in information.

## Objective
Add a Step model and use it in the Build model. With a separate field for each key in a Step, we can better prevent race conditions.

_Note: Making `command` optional since the launcher setup step doesn't have any commands. Making `id` and `buildId` optional for backwards compatibility._

## Related links
This bug is related to changes made for https://github.com/screwdriver-cd/screwdriver/issues/1114


models/step.js
```diff
@@ -0,0 +1,123 @@
+'use strict';
+
+const Joi = require('joi');
+const mutate = require('../lib/mutate');
+
+const MODEL = {
+    id: Joi
```
jithin1987: is this field required ? `name` & `buildId` should be enough to identify it.

minz1027: This is what it's going to put into postgres. Every model has an ID field. https://github.com/screwdriver-cd/data-schema/blob/master/models/banner.js#L7

PR #272, opened 2018-08-21
fix: Do not allow dangerous shell metacharacters in image name

tkyi:
To avoid command injection attack in shell commands, we should reject user-supplied input if it includes dangerous shell metacharacters.


tkyi:
@d2lam Added a test!

PR #273, opened 2018-08-28
[1] fix(1220): Add scmInfo to scm module functions

catto:
# Context
I found that there are 9 or more `lookupScmUri` calls in models during a single PR event even if it is rarely changed by renaming a repository or switching branches. It should be cached in each pipeline instance.

# Objective
I added an optional `scmInfo` schema to `getFile` and `decorateUrl` function to accept cached scmInfo object provided by pipeline model.
It doesn't change current behavior because of optional keys.

# References
screwdriver-cd/screwdriver#1220

### DIFF
--- a/plugins/scm.js
+++ b/plugins/scm.js
@@ -17,7 +17,7 @@ const scmInfo = Joi.object({
     host: Joi.string().required(),
     owner: Joi.string().required(),
     repo: Joi.string().required()
-});
+}).optional();
 const sha = Joi.reach(models.build.base, 'sha').required();
 const token = Joi.reach(models.user.base, 'token').required();
 const type = Joi.reach(core.scm.hook, 'type').required();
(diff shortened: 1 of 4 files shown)


plugins/scm.js
```diff
@@ -12,6 +12,12 @@ const pipelineId = Joi.reach(models.pipeline.base, 'id').optional();
 const prNum = Joi.reach(core.scm.hook, 'prNum').allow(null).optional();
 const scmContext = Joi.reach(models.pipeline.base, 'scmContext').optional();
 const scmUri = Joi.reach(models.pipeline.base, 'scmUri').required();
+const scmInfo = Joi.object({
```
minz1027: Why not just do `.optional()` here? Also, can we have a test for this change?

catto: @minz1027 Thanks for the comment! I have fixed it

PR #276, opened 2018-08-31
feat(1220): Allow passing scmRepo to some functions

catto:
# Context
I would like to pass cached scmInfo data fetched via `lookupScmUri` but it's a private function. #273 

# Objective
Pass scmRepo instead of scmInfo since lookupScmUri is private function. We can get `scmRepo` from pipeline model (database).

# References
screwdriver-cd/screwdriver#1220

### DIFF
--- a/plugins/scm.js
+++ b/plugins/scm.js
@@ -3,6 +3,7 @@
 const core = require('../core');
 const Joi = require('joi');
 const models = require('../models');
+const Scm = require('../core/scm');
 
 const buildStatus = Joi.reach(models.build.base, 'status').required();
 const checkoutUrl = Joi.reach(models.pipeline.create, 'checkoutUrl').required();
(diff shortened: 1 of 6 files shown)


wdstar:
We should rename test data files.
- scm.getPermissionsWithScmRepo.yaml
- scm.getCommitShaWithScmRepo.yaml

yuichi10:
It looks like affect many code, can we merge this code?

catto:
@yuichi10 It includes only `optional` argument changes. This PR replaces `scmInfo`  but it is added very recently (#273) and there is no running environment because of broken design. Also I'm going to bump minor version.

catto:
@wdstar I forgot to add test data. I fixed it and now files seem to be renamed.

catto:
Following is #277

PR #278, opened 2018-09-06
fix(880): Add search to pagination and datastore scan schemas

tkyi:
## Context
Since we're adding pagination for pipelines, we can no longer have search functionality stored in the UI. We need to move this logic to the datastore.

## Objective
This PR adds search schemas for the datastore and pagination API calls.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/880

### DIFF
--- a/plugins/datastore.js
+++ b/plugins/datastore.js
@@ -24,6 +24,10 @@ const SCHEMA_SCAN = Joi.object().keys({
         count: Joi.number().integer().positive().required(),
         page: Joi.number().integer().positive().required()
     }),
+    search: Joi.object().keys({
+        field: Joi.string().max(100).required(),
+        term: Joi.string().max(200).required()
+    }),
     sort: Joi.string().lowercase().valid(['ascending', 'descending']).default('descending'),
     sortBy: Joi.string().max(100)
 });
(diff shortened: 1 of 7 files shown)


plugins/datastore.js
```diff
@@ -24,6 +24,10 @@ const SCHEMA_SCAN = Joi.object().keys({
         count: Joi.number().integer().positive().required(),
         page: Joi.number().integer().positive().required()
     }),
+    search: Joi.object().keys({
+        field: Joi.string().max(100).required(),
+        term: Joi.string().max(200).required()
```
d2lam: `field` and `term` sound similar to me. What is `term`? Is it the value to search for? Would `value` be less confusing?

tkyi: I guess search `term` or search `keyword` captures what I mean the most (https://www.opentracker.net/article/choosing-search-terms), but when those words are compared with `field`, it is confusing.

Would `key` and `value` be clearer? `field` and ` value`?

d2lam: I think `key/value` or `field/keyword` would make sense.
I'm leaning toward `key/value` because it's easy to understand...

tkyi: Changed to `field/keyword` 👍

PR #284, opened 2018-09-27
[DNM] fix(1241): Add branch as optional event field

tkyi:
## Context
Since branch-specific triggers have been added, it would be nice if the UI would also reflect when a trigger was done by a branch-specific trigger. However, the UI doesn't know what branch a job was run on.

## Objective
This PR adds `branch` as an optional field for `events`. This will be needed in the UI.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1241


tkyi:
Don't need this PR anymore since the `startFrom` will include the branch

PR #294, opened 2018-11-16
fix(1319): allow buildClusterName in executor.start

minz1027:
allow optional buildClusterName in executor.start 

https://github.com/screwdriver-cd/screwdriver/issues/1319


tkyi:
Probably need buildCluster name in executor stop too

minz1027:
good catch! addressed :D

PR #296, opened 2018-11-22
[1] feat(1010): Add prChain flag for PR workflow.

sakka2:
## Context
I would like to add a prChain flag to the base config in order to allow PRs triggering subsequent jobs.

## Objective
- add `SCHEMA_PR_CHAIN` into `SCHEMA_CONFIG`
`SCHEMA_PR_CHAIN` should be boolean and default value is `false` if it is not set.

## References for the proposal
https://github.com/screwdriver-cd/screwdriver/issues/1010#issuecomment-436195792

## Related PRs
https://github.com/screwdriver-cd/config-parser/pull/81
https://github.com/screwdriver-cd/launcher/pull/241


models/job.js
```diff
@@ -18,6 +18,11 @@ const MODEL = {
 
     permutations: validator.jobPermutations,
 
+    prNum: Joi
```
tk3fftk: If a job is not a PR job, `prNum` may be `undefined` but it should cause validation error.

wdstar: `optional()` is not necessarily needed.
> Marks a key as optional which will allow undefined as values. Used to annotate the schema for readability as all keys are optional by default.

https://github.com/hapijs/joi/blob/v14.3.0/API.md#anyoptional

tk3fftk: I see.
I concerned about cases of `prNum: undefined` or `prNum: null` pattern. It's same as `configPipelineId` on 
https://github.com/screwdriver-cd/data-schema/blob/master/models/pipeline.js#L60-L62

wdstar: We should use `default(null)`?

tk3fftk: Default `optional()` might work.

sakka2:
I added `prChain` to `validator` and `pipeline`

models/job.js
sakka2: I deleted `prNum` for this time.
I will be added in Second Pass. 
https://github.com/screwdriver-cd/screwdriver/issues/1010#issuecomment-441493555

PR #298, opened 2018-11-30
feat(1271): extend query options

DekusDenial:
extending query options to datastore
related to: https://github.com/screwdriver-cd/screwdriver/issues/1271

### DIFF
--- a/models/command.js
+++ b/models/command.js
@@ -10,25 +10,23 @@ const MODEL = {
         .number().integer().positive()
         .description('Identifier of this command')
         .example(123345),
-
+    namespace: Command.namespace,
+    version: Command.version,
+    description: Command.description,
+    maintainer: Command.maintainer,
+    format: Command.format,
+    habitat: Command.habitat,
+    docker: Command.docker,
+    binary: Command.binary,
+    name: Command.name,
+    pipelineId,
     createTime: Joi
         .string()
         .isoDate()
         .max(32)
         .description('When this command was created')
         .example('2038-01-19T03:14:08.131Z'),
-
-    namespace: Command.namespace,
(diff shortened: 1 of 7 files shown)


models/template.js
```diff
@@ -34,8 +24,15 @@ const MODEL = {
     version: Template.version,
     description: Template.description,
     maintainer: Template.maintainer,
+    pipelineId,
+    namespace: Template.namespace,
     images: Template.images,
-    pipelineId
+    createTime: Joi
```
tkyi: any reason why the order of these fields were changed?

plugins/datastore.js
```diff
@@ -33,7 +33,9 @@ const SCHEMA_SCAN = Joi.object().keys({
         keyword: Joi.string().max(200).required()
     }),
     sort: Joi.string().lowercase().valid(['ascending', 'descending']).default('descending'),
-    sortBy: Joi.string().max(100)
+    sortBy: Joi.string().max(100),
+    exclude: Joi.array().items(Joi.string()),
```
tkyi: might be better to put max as well, as in `Joi.array().items(Joi.string().max(100)),`

models/command.js
```diff
@@ -10,25 +10,23 @@ const MODEL = {
...
     version: Command.version,
     description: Command.description,
-    usage: Command.usage,
     maintainer: Command.maintainer,
     format: Command.format,
     habitat: Command.habitat,
     docker: Command.docker,
     binary: Command.binary,
-    pipelineId
+    name: Command.name,
```
tkyi: any reason to change the order of fields?

DekusDenial: same as above

PR #300, opened 2018-12-12
[1] feat(1270): add stats to build model

d2lam:
Allow optional build field in executor start and add stats to build model

Related: https://github.com/screwdriver-cd/screwdriver/issues/1270

### DIFF
--- a/models/build.js
+++ b/models/build.js
@@ -110,6 +110,7 @@ const MODEL = {
             hostname: Joi.string()
                 .description('Where this build is run')
         })
+        .unknown(true) // allow other fields
         .description('Stats for this build'),
 
     buildClusterName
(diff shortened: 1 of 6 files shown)


models/build.js
```diff
@@ -100,6 +100,18 @@ const MODEL = {
         .description('Status message to describe status of the build')
         .example('Build failed due to infrastructure error'),
 
+    stats: Joi
```
jithin1987: can the keys be dynamic ? If tomorrow we want to track another build stat, can we do that without a code change ?

jithin1987: For example if custom executors want to track metrics specific to them, say executor iOS one wanting to track VM setup time, executor  hyperd tracking VM creation time.

d2lam: yeah we can make it just `object`

d2lam: Use `unknown` option instead so that for defined fields like `queueTime` it will check that the it's `isoDate`. This option will still allow adding other fields.

PR #301, opened 2018-12-15
feat(1355): Add PR comment schema [1]

tkyi:
## Context
It would be nice for users to be able to comment back on PRs through a build with relevant metadata.

## Objective
This PR adds a schema for adding a PR comment.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1355

### DIFF
--- a/plugins/scm.js
+++ b/plugins/scm.js
@@ -72,6 +72,14 @@ const GET_COMMIT_SHA = Joi.object().keys({
     scmRepo: Scm.repo.optional()
 }).required();
 
+const ADD_PR_COMMENT = Joi.object().keys({
+    scmUri,
+    token,
+    prNum,
+    comment: Joi.string().required(),
+    scmContext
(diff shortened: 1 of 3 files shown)


plugins/scm.js
```diff
@@ -72,6 +72,14 @@ const GET_COMMIT_SHA = Joi.object().keys({
     scmRepo: Scm.repo.optional()
 }).required();
 
+const ADD_PR_COMMENT = Joi.object().keys({
+    scmUri,
+    token,
+    prNum,
```
minz1027: this is optional when it's defined: https://github.com/screwdriver-cd/data-schema/blob/8ba40a1be92babc54d73bedb5aea5a6cad932b51/plugins/scm.js#L13

is it the intention here to make it optional?

tkyi: no, I'll change it to make it required

PR #302, opened 2019-01-09
feat(1419): extend job schema for pr job info

DekusDenial:
added more info the PR and PR job

related: https://github.com/screwdriver-cd/screwdriver/issues/1419

### DIFF
--- a/core/scm.js
+++ b/core/scm.js
@@ -72,14 +72,23 @@ const SCHEMA_COMMIT = Joi.object().keys({
         .uri()
         .required()
         .label('Link to commit')
-        .example('https://github.com/scredriver-cd/screwdriver/commit/8843d7f92416211de')
+        .example('https://github.com/screwdriver-cd/screwdriver/commit/8843d7f92416211de')
 }).label('SCM Commit');
 
 const SCHEMA_PR = Joi.object().keys({
     url: Joi.string()
         .uri()
         .label('Link to PR')
-        .example('https://git.ouroath.com/MAILSERVICES/gdpr-event-processor/pull/1')
+        .example('https://github.com/screwdriver-cd/screwdriver/pull/1'),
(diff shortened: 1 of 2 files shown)


core/scm.js
```diff
@@ -72,14 +73,24 @@ const SCHEMA_COMMIT = Joi.object().keys({
...
 const SCHEMA_PR = Joi.object().keys({
     url: Joi.string()
         .uri()
         .label('Link to PR')
-        .example('https://git.ouroath.com/MAILSERVICES/gdpr-event-processor/pull/1')
+        .example('https://github.com/screwdriver-cd/screwdriver/pull/1'),
+    title: Joi.string()
+        .max(512)
+        .label('Title of the pull request'),
+    createTime: Joi.date().iso()
```
tkyi: (just wondering) What's the difference between `Joi.date().iso()` vs `Joi.string().isoDate()`? Since we seem to use the second one everywhere else

core/scm.js
```diff
@@ -72,14 +73,24 @@ const SCHEMA_COMMIT = Joi.object().keys({
...
+        .label('Creation Time of the pull request')
+        .example('2018-10-10T21:35:31Z'),
+    username: Joi.string()
+        .max(39)
```
tkyi: Should we change the `user` username field to match? Probably not urgent though
https://github.com/screwdriver-cd/data-schema/blob/master/models/user.js#L14

core/scm.js
DekusDenial: The former is the "base check" while the latter does more to ensure iso compatibility. For external API, i will prefer using the generic type check.

core/scm.js
DekusDenial: the more i think about it, i should remove this max since the definition of max varies per scm provider

PR #325, opened 2019-03-12
fix(1010): [4] delete prChain from config validator.

wdstar:
## Context

We moves prChain flag from the top level setting `prChain` to the pipeline level annotation `chainPR`.

## Objectives

This PR deletes the old `prChain` flag from the config validator.
But it preserves the `pipeline.prChain` property for the current Models and UI implementations.

## References

- https://github.com/screwdriver-cd/screwdriver/issues/1010#issuecomment-466188824
- https://github.com/screwdriver-cd/screwdriver/pull/1532
- https://github.com/screwdriver-cd/models/pull/344


config/base.js
```diff
@@ -59,7 +58,7 @@ const SCHEMA_CONFIG = Joi.object()
 module.exports = {
     jobs: SCHEMA_JOBS,
     shared: SCHEMA_SHARED,
-    prChain: SCHEMA_PR_CHAIN,
+    prChain: SCHEMA_PR_CHAIN, // This is still used by pipeline schema.
```
s-yoshika: Is this  still needed?

wdstar: Yes, for the current models and UI (via pipeline API).
Please refer to the following code.
- https://github.com/screwdriver-cd/models/pull/344/files#diff-679e90b5b15ab1b73ddb7ed2377e7413R637

s-yoshika: I may be mistaken, but I think the definition of pipeline model is [here](https://github.com/screwdriver-cd/data-schema/blob/master/models/pipeline.js#L67), doesn't it?

wdstar: Yes, the definition of pipeline model refers to this `Base.prChain` here.

s-yoshika: Oh.. I overlooked :bow:

PR #333, opened 2019-04-26
feat(1455): Add rootDir for pipeline create and update (1)

tkyi:
## Context
Would be nice if users could put screwdriver.yaml not at root.

## Objective
This PR adds `rootDir` to pipeline create/update, `scmRepo`, and scm method `parseUrl`

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1455

### DIFF
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -31,7 +31,8 @@ const MODEL = {
     scmUri: Joi
         .string().regex(Regex.SCM_URI).max(128)
         .description('Unique identifier for the application')
-        .example('github.com:123456:master'),
+        .example('github.com:123456:master')
+        .example('github.com:123456:master:src/app/component'),
 
     scmContext: Joi
         .string().max(128)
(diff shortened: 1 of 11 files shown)


models/pipeline.js
```diff
@@ -101,15 +104,19 @@ module.exports = {
      * @property create
      * @type {Joi}
      */
-    create: Joi.object(CHECKOUT_URL).label('Create Pipeline'),
+    create: Joi.object(mutate(CREATE_MODEL, [], [
+        'checkoutUrl', 'rootDir'
```
minz1027: for create, `checkoutUrl` should be required

plugins/scm.js
```diff
@@ -151,6 +153,7 @@ const DECORATE_AUTHOR = Joi.object().keys({
 
 const PARSE_URL = Joi.object().keys({
     checkoutUrl,
+    rootDir,
```
minz1027: we also need `rootDir` in for GET_CHECKOUR_COMMAND

models/pipeline.js
tkyi: oops, fixed

PR #337, opened 2019-05-10
fix(823): [1] Add some release fields for SCHEMA_HOOK

kkokufud:
## Context
SD can execute release/tag trigger, but it can't provide release/tag info(ex. name).
SD user want to get release/tag info.

## Objective
Add some release fields to SCHEMA_HOOK.
Please see reference to confirm details.

## References
https://github.com/screwdriver-cd/screwdriver/issues/823#issuecomment-490014132

### DIFF
--- a/core/scm.js
+++ b/core/scm.js
@@ -187,7 +187,24 @@ const SCHEMA_HOOK = Joi.object().keys({
         .required()
         .label('Type of the event'),
 
-    username: Joi.reach(SCHEMA_USER, 'username')
+    username: Joi.reach(SCHEMA_USER, 'username'),
+
+    releaseId: Joi.number()
+        .integer()
+        .positive()
+        .optional()
+        .label('Release id'),
+
+    releaseName: Joi.string()
+        .allow('')
+        .optional()
+        .label('Name of the event'),
+
+    releaseAuthor: Joi.string()
+        .allow('')
+        .optional()
+        .label('Author of the event')
+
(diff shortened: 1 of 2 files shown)


core/scm.js
```diff
@@ -187,7 +187,24 @@ const SCHEMA_HOOK = Joi.object().keys({
...
-    username: Joi.reach(SCHEMA_USER, 'username')
+    username: Joi.reach(SCHEMA_USER, 'username'),
+
+    releaseId: Joi.number()
+        .integer()
+        .positive()
+        .optional()
+        .label('Release id'),
+
+    releaseName: Joi.string()
```
tk3fftk: Does "release" include "tag"? If so, I think these field names are confusable.

tk3fftk: I found it does not include "tag" in other pr.

core/scm.js
```diff
@@ -187,7 +187,24 @@ const SCHEMA_HOOK = Joi.object().keys({
         .required()
         .label('Type of the event'),
 
-    username: Joi.reach(SCHEMA_USER, 'username')
+    username: Joi.reach(SCHEMA_USER, 'username'),
+
+    releaseId: Joi.number()
+        .integer()
```
wdstar: We should use `string`. Maybe another SCM's release IDs include non-numbers.

kkokufud: Thank you your comments. I changed type of releaseId number to string.

PR #340, opened 2019-06-20
fix: add "trusted" column to template and command

DekusDenial:
## Context

To mark certain template and command as trusted, we need to modify the existing schema to include additional column on the db table

## Objective

add `trusted` column to template's and command's schema

## References

https://github.com/screwdriver-cd/screwdriver/issues/1427

## License


DekusDenial:
@minz1027 I was almost afraid that u wouldn't ask!

PR #344, opened 2019-08-12
feat(947): Added a new type field for collection schema

Code-Beast:
## Context
Would be nice if there's a default collection where a user can find all pipelines they create.

## Objective
This pull requests adds a new "type" field to collection schema

## References
Related to [947](https://github.com/screwdriver-cd/screwdriver/issues/947)

## License

### DIFF
--- a/models/collection.js
+++ b/models/collection.js
@@ -41,7 +41,13 @@ const MODEL = {
 
     pipelineIds: Joi
         .array()
-        .items(Joi.number().integer().positive())
+        .items(Joi.number().integer().positive()),
+
+    type: Joi
+        .string()
+        .max(32)
+        .description('Collection type')
+        .example('default')
 };
 const GET_MODEL = Object.assign({}, MODEL, { pipelines: PIPELINES_MODEL });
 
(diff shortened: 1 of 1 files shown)


models/collection.js
```diff
@@ -77,7 +85,8 @@ module.exports = {
         'id',
         'name',
         'pipelineIds',
-        'userId'
+        'userId',
+        'type'
```
tkyi: let's not make the `type` field required
just make it optional so we don't need to do a DB migration

models/collection.js
```diff
@@ -64,7 +71,8 @@ module.exports = {
         'id',
         'name',
         'pipelineIds',
-        'pipelines'
+        'pipelines',
+        'type'
```
tkyi: let's not make the `type` field required
just make it optional so we don't need to do a DB migration

models/collection.js
```diff
@@ -41,7 +41,14 @@ const MODEL = {
 
     pipelineIds: Joi
         .array()
-        .items(Joi.number().integer().positive())
+        .items(Joi.number().integer().positive()),
+
+    type: Joi
+        .string()
+        .max(32)
+        .valid(['default', 'normal'])
```
tkyi: let'd allow empty

models/collection.js
Code-Beast: I see, thanks for reminding

PR #348, opened 2019-09-10
fix: add index

parthasl:
## Context

Improve DB query performance by using an index scan

## Objective

Added index based on the data access pattern from the database metrics

## References

pipelineId and state columns used in where clause query of the jobs table will fire a single index scan

## License

### DIFF
--- a/models/job.js
+++ b/models/job.js
@@ -141,5 +141,5 @@ module.exports = {
      * @property indexes
      * @type {Array}
      */
-    indexes: [{ fields: ['pipelineId'] }, { fields: ['state'] }]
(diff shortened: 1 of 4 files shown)


models/job.js
```diff
@@ -141,5 +141,5 @@ module.exports = {
      * @property indexes
      * @type {Array}
      */
-    indexes: [{ fields: ['pipelineId'] }, { fields: ['state'] }]
+    indexes: [{ fields: ['pipelineId', 'state'] }, { fields: ['state'] }]
```
adong: ` { fields: ['state'] }` do we still need the 2nd field?

parthasl: yes, this is needed, as it's second in the index; this field will not be used when used separately in query;

PR #351, opened 2019-09-16
fix(1747): adding commit authors to hook schema

klu909:
## Context

Currently there's a mismatch in v3 and v4 ci skip logic. To change v4 ci skip logic to match v3 we need an additional field.

## Objective

This PR add an optional field to hook schema

## References

https://github.com/screwdriver-cd/screwdriver/issues/1747

## License

### DIFF
--- a/core/scm.js
+++ b/core/scm.js
@@ -192,6 +192,11 @@ const SCHEMA_HOOK = Joi.object().keys({
 
     username: Joi.reach(SCHEMA_USER, 'username'),
 
+    commitAuthors: Joi.array()
+        .items(Joi.string())
+        .optional()
+        .label('Commit authors'),
+
     releaseId: Joi.string()
         .allow('')
         .optional()
--- a/test/core/scm.test.js
+++ b/test/core/scm.test.js
@@ -64,6 +64,14 @@ describe('scm core', () => {
             assert.isNull(validate('scm.hook.push.yaml', core.scm.hook).error);
         });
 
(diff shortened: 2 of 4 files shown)


tkyi:
Can you add a test?

test/core/scm.test.js
```diff
@@ -64,6 +64,10 @@ describe('scm core', () => {
             assert.isNull(validate('scm.hook.push.yaml', core.scm.hook).error);
         });
 
+        it('fails the push hook', () => {
+            assert.isNotNull(validate('scm.hook.push.commiters.yaml', core.scm.hook).error);
```
jithin1987: can you call this file `cm.hook.push.invalid-commiters.yaml`

core/scm.js
```diff
@@ -192,6 +192,10 @@ const SCHEMA_HOOK = Joi.object().keys({
 
     username: Joi.reach(SCHEMA_USER, 'username'),
 
+    commitAuthors: Joi.array()
```
tkyi: I think it'd be safer to specify what kind of things can be in the array, like https://github.com/screwdriver-cd/data-schema/blob/master/models/template.js#L15

d2lam: I think this should be `Joi.array().items(Joi.string())`

klu909: change done, is there something else that I should change?

klu909: Thanks for info, I have changed it

PR #356, opened 2019-09-30
fix(1763): add migration files to extend column for object to mediumtext

s-yoshika:
## Context
In [this PR](https://github.com/screwdriver-cd/datastore-sequelize/pull/58), we changed the data type of columns which are defined as `object` in`data-schema` from `TEXT` to  `MEDIUMTEXT`.

## Objective
This add migration files for above changes.
I confirmed that the columns defined as `object` in datas-schema are following.
* `events.meta`
* `builds.meta`
* `builds.parameters`
* `pipelines.admins`
https://github.com/screwdriver-cd/data-schema/search?q=path%3Amodels+object&unscoped_q=path%3Amodels+object
## Migration issue due to this change
This also changes `initdb-*` files.
Although the migration files are executed in order by filename,
it doesn't assure that `initdb-*` files are executed first.
https://stackoverflow.com/questions/33429383/sequelize-migration-order-of-execution

So this PR appends date prefix into all migration files to guarantee the execution order.

But there is issue for this changes.
Any migrations which already executed in your DB are recorded in their own table(`SequelizeMeta`) like this.
```
mysql> select * from SequelizeMeta;
+-------------------------+
| name                    |
+-------------------------+
| initdb-banners.js       |
| initdb-buildClusters.js |
| initdb-builds.js        |
```
It means that re-executes `initdb-*` even if the migration files are executed already,  because these file names will be changed with date prefix after this PR is merged.

Please execute like following sql in your DB if your cluster already executed migrations for `initdb-*`.
```
UPDATE SequelizeMeta SET name = CONCAT('20190919-', name);
```

Or if there are another good ways, please comment.

## References
https://github.com/screwdriver-cd/screwdriver/issues/1763

## License


migrations/20190919-initdb-builds.js
```diff
@@ -52,10 +52,10 @@ module.exports = {
                     type: Sequelize.TEXT
                 },
                 parameters: {
-                    type: Sequelize.TEXT
+                    type: Sequelize.TEXT('medium')
```
s-yoshika: if the dialect doesn't support, same as https://github.com/screwdriver-cd/datastore-sequelize/pull/58/files#r328474936

sugarnaoming:
[ask]
Is this prefix means execution `initdb-*` files first ?
Is it necessary to guarantee the execution order in `initdb-*` files ?

s-yoshika:
@sugarnaoming 

> Is this prefix means execution initdb-* files first ?

Yes.

> Is it necessary to guarantee the execution order in initdb-* files ?

Each `initdb-*` files create each tables in Screwdriver database by  `CREATE TABLE` query, 
and other migration files migrate those tables by the query like `ALTER COLUMN` which expects target tables already exists.
So I think `initdb-*` files needs to be executed first than any other files.

s-yoshika:
> Is it necessary to guarantee the execution order in initdb-* files ?

Sorry I might missunderstanding your quetion.

Maybe No.
I think each `initdb-*` files just create each tables and these operations are exclusive.

sugarnaoming:
Thank you.
I understand.

sugarnaoming:
LGTM 👍

parthasl:
@s-yoshika understood file name change. Init captures a snapshot of the current DB schema, appreciate new migration files for any go-forward schema changes.

s-yoshika:
@parthasl I thought `init` files ensure that any necessary setup for DB is completed. And it's execution is skipped if it is already executed by old query. So changing init files don't affect existing cluster. But it’s certainly redundant for new cluster. It may needless.

tk3fftk:
@parthasl 
Do we have any policies which is updating `init` or just creating new files?
Th both ways are looks ok.

parthasl:
> @parthasl
> Do we have any policies which is updating `init` or just creating new files?
> Th both ways are looks ok.

Init files match existing schema in production. for any new changes, create new files with datetime in the filename. Yes, if init files are already updated in db, it will skip.

catto:
`initdb-` should contain just a snapshot of schema in database at the time when `initdb-` files were created. Let's create only new migration files to change the schema like we do git commit to change files.

tk3fftk:
> Let's create only new migration files to change the schema like we do git commit to change files.

It looks a better way. I'll remove `init` file changes.

migrations/20190930-extend_objects_to_mediumtext-builds.js
```diff
@@ -0,0 +1,22 @@
...
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        await queryInterface.sequelize.transaction(async (transaction) => {
+            await queryInterface.changeColumn(table, 'meta', Sequelize.TEXT('medium'),
+                { transaction }
+            );
+        });
+
+        await queryInterface.sequelize.transaction(async (transaction) => {
+            await queryInterface.changeColumn(table, 'parameters', Sequelize.TEXT('medium'),
```
jithin1987: it looks like we are not using this column at all. Better to remove this in a future PR. I will drop a separate issue for this.

jithin1987: https://github.com/screwdriver-cd/screwdriver/issues/1774

PR #360, opened 2019-10-08
fix(1675): Add causeMessage to executor start [1]

tkyi:
## Context

Sometimes users want to override their freeze window quickly without changing the screwdriver.yaml.

## Objective

This PR adds a field `causeMessage` to executor start. Users can pass in `[force start]` keywords (similar to [skip ci]) to force a build to start during a freeze window.

_Note: Not sure if `force start` is the best keyword phrase. Open to other ideas: `override frozen`, `ci` (as the opposite of `skip ci`?), etc._

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1675

## License

### DIFF
--- a/plugins/executor.js
+++ b/plugins/executor.js
@@ -6,6 +6,7 @@ const Joi = require('joi');
 const models = require('../models');
 const buildId = Joi.reach(models.build.base, 'id').required();
 const eventId = Joi.reach(models.event.base, 'id');
+const causeMessage = Joi.reach(models.event.base, 'causeMessage');
 const jobId = Joi.reach(models.job.base, 'id');
 const jobName = Joi.reach(models.job.base, 'name');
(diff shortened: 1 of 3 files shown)


jithin1987:
I like `force start`, more generic and conveys intention,

plugins/executor.js
```diff
@@ -18,6 +19,7 @@ const SCHEMA_PIPELINE = Joi.object().keys({
 
 const SCHEMA_START = Joi.object().keys({
     build: Joi.object(),
+    causeMessage,
```
jithin1987: does this need DB change ?

tkyi: it's not changing a model, so I don't think so
just adding a param to what the executor is expecting for the `start()`

PR #361, opened 2019-10-16
feat(1710): allow external join, add parentBuilds field. BREAKING CHANGE: allow external AND

d2lam:
Related: https://github.com/screwdriver-cd/screwdriver/issues/1710
Please DNM yet


test/data/build.get.yaml
```diff
@@ -10,5 +10,14 @@ stats:
   blockedStartTime: '2017-01-06T01:49:50.384359267Z'
   queueEnterTime: '2017-01-06T01:49:50.384359267Z'
   imagePullStartTime: '2017-01-06T02:49:50.384359267Z'
+parentBuilds:
+  111:
+    eventId: 1,
```
tkyi: what if there's a job named `eventId`?

d2lam: That’s a really weird job name but totally possible. Maybe we can put all the jobs under keyword “jobs”. Then we can access by “111.jobs.A”

d2lam: will need to change API logic (https://github.com/screwdriver-cd/screwdriver/pull/1805) to check build status based on `111.jobs.A` instead of `111.A`

d2lam:
LGTM me @tkyi

PR #366, opened 2019-12-04
fix: increasing SourcePath char limit to 1024

abhinvsinh:
## Context

Currently the filename limit for SourcePath is set to 100 and thus validation fails when we try to use a filename which is greater than 100 characters.  Requesting an increase to 1024 characters.
## Objective

PR fixes the filename character limit

## License


adong:
@abhinvsinh Just FYI, [Screwdriver](https://github.com/screwdriver-cd) uses [semantic release](https://github.com/semantic-release/semantic-release), which will auto bump the version during CICD build.

https://docs.screwdriver.cd/about/contributing/
image

PR #368, opened 2019-12-16
feat(1881): build cluster to support multple scm conexts

parthasl:
## Context

Build cluster supports only one SCM context, which leads to error "This pipeline is not authorized to use this build cluster." if the same build cluster is used for different SCM context.

## Objective

This PR is intended for a build cluster to support multiple SCM contexts

## References

[1881](https://github.com/screwdriver-cd/screwdriver/issues/1881)

## License

### DIFF
--- a/migrations/20191214-rename_scmContext-buildClusters.js
+++ b/migrations/20191214-rename_scmContext-buildClusters.js
@@ -1,0 +1,32 @@
+/* eslint-disable new-cap */
+
+'use strict';
+
+const prefix = process.env.DATASTORE_SEQUELIZE_PREFIX || '';
+const table = `${prefix}buildClusters`;
+
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        try {
+            await queryInterface.sequelize.query('SET SQL_MODE = CONCAT(@@SQL_MODE, ' +
+                '\',ANSI_QUOTES\')');
+            // eslint-disable-next-line no-empty
+        } catch (e) {}
+
+        await queryInterface.sequelize.transaction(async (transaction) => {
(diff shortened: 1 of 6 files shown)


test/data/buildCluster.create.yaml
```diff
@@ -1,5 +1,6 @@
 # BuildCluster create example
 name: iOS
+scmContexts: ['github:github.com', 'github:git.com']
```
DekusDenial: Is `git.com` a typo?

migrations/20191214-rename_scmContext-buildClusters.js
```diff
@@ -0,0 +1,32 @@
...
+'use strict';
+
+const prefix = process.env.DATASTORE_SEQUELIZE_PREFIX || '';
+const table = `${prefix}buildClusters`;
+
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        try {
+            await queryInterface.sequelize.query('SET SQL_MODE = CONCAT(@@SQL_MODE, ' +
+                '\',ANSI_QUOTES\')');
```
DekusDenial: Will Postgres ignore these?

test/data/buildCluster.create.yaml
parthasl: no, just gave some random scm context

migrations/20191214-rename_scmContext-buildClusters.js
parthasl: yes, it will ignore

test/data/buildCluster.create.yaml
parthasl: updated

PR #379, opened 2020-01-21
fix(1865): [3] Remove steps column in builds table

sakka2:
## Context
As mentioned here screwdriver-cd/screwdriver#1865 , builds.steps and steps.command are both same data.
And these steps are defined by user's yaml and it tends to have a big size.

## Objective
Create migration file which removes steps column in builds table, because we fixed not to use steps column.

## References
- https://github.com/screwdriver-cd/models/pull/423
- https://github.com/screwdriver-cd/screwdriver/issues/1865

## License

### DIFF
--- a/migrations/20200114-del-column-builds-steps.js
+++ b/migrations/20200114-del-column-builds-steps.js
@@ -13,3 +13,4 @@ module.exports = {
         });
     }
 };
+
--- a/migrations/20200114-del-column-builds-steps.js
+++ b/migrations/20200114-del-column-builds-steps.js
@@ -8,9 +8,10 @@ const table = `${prefix}builds`;
 module.exports = {
     // eslint-disable-next-line no-unused-vars
     up: async (queryInterface, Sequelize) => {
-        await queryInterface.sequelize.transaction(async (transaction) => {
-            await queryInterface.removeColumn(table, 'steps', { transaction });
-        });
+        try {
+            await queryInterface.removeColumn(table, 'steps');
+            // eslint-disable-next-line no-empty
+        } catch (e) {}
     }
 };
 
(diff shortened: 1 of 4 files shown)


tk3fftk:
Is it enough only to add migration file?

migrations/20200114-del-column-builds-steps.js
```diff
@@ -0,0 +1,16 @@
...
+
+'use strict';
+
+const prefix = process.env.DATASTORE_SEQUELIZE_PREFIX || '';
+const table = `${prefix}builds`;
+
+module.exports = {
+    // eslint-disable-next-line no-unused-vars
+    up: async (queryInterface, Sequelize) => {
+        await queryInterface.sequelize.transaction(async (transaction) => {
```
parthasl: no need for a single transaction

sugarnaoming: I think it is better to be able to roll back if you fail on the way.

sugarnaoming:
> Don't we need to remove this from `build` model ? [/models/build.js@`master`#L95-L97](https://github.com/screwdriver-cd/data-schema/blob/master/models/build.js#L95-L97)

I see.
It is miss commit, so reverted.

models/build.js
```diff
@@ -92,10 +92,6 @@ const MODEL = {
         .default({})
         .description('Key=>Value information from the build itself'),
 
-    steps: Joi
```
tk3fftk: Can we remove `steps` from model? I'm worried this makes validation failure like that line.
https://github.com/screwdriver-cd/screwdriver/blob/b2aa7c77fdd4470e87e172592489575009151ea6/plugins/builds/get.js#L51

sugarnaoming: I see.
this fixed.

migrations/20200114-del-column-builds-steps.js
parthasl: u can keep it - but no use, as there is only one statement getting executed; it will be useful for multiple statements

kkisic: I see.
I fixed it.

PR #381, opened 2020-01-30
feat(1091): add template id to jobs and builds [1]

InderH:
## Context

Need to add template id to job schema so metrics about the number of jobs using some template can be aggregated.

## Objective

This PR adds template id to the job schema under the field templateId.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1091

## License


config/job.js
```diff
@@ -118,6 +118,10 @@ const SCHEMA_STEPS_NO_DUPS = Joi.array().items(SCHEMA_STEP).min(1).unique((a, b)
     return Object.keys(a).some(key => b[key]);
 });
 const SCHEMA_TEMPLATE = Joi.string().regex(Regex.FULL_TEMPLATE_NAME);
+const SCHEMA_TEMPLATEID = Joi
```
jithine: this should be marked as options/`allow null` else existing records will break

tkyi:
I had a couple questions:
1. Will we also store `templateId` in builds?
2. Since a unique template is determined by `namespace`, `name`, and `version`,
  a. are we planning to return popularity information based on a specific template version?
  b. or based on template namespace and name?

InderH:
> I had a couple questions:
> 
> 1. Will we also store `templateId` in builds?
> 2. Since a unique template is determined by `namespace`, `name`, and `version`,
>    a. are we planning to return popularity information based on a specific template version?
>    b. or based on template namespace and name?

1. Yes, templateId will also be stored in builds
2. the api will return metrics for each template version

PR #384, opened 2020-02-17
fix: templateId is DOUBLE

yoshwata:
## Context

There is some code that should be DOUBLE not INTEGER.

## Objective

Migration works correctly.

## References

https://github.com/screwdriver-cd/data-schema/blob/dde2d906e399ce15e87b1e1808cf908e68613705/models/job.js#L70

https://github.com/screwdriver-cd/data-schema/blob/dde2d906e399ce15e87b1e1808cf908e68613705/models/build.js#L132

https://github.com/screwdriver-cd/datastore-sequelize/blob/8fd9efbbcce52d82be0867c15303f9bd2b0a4180/index.js#L116

## License

### DIFF
--- a/migrations/20200203-add-column_templateId_to_jobs.js
+++ b/migrations/20200203-add-column_templateId_to_jobs.js
@@ -9,7 +9,7 @@ module.exports = {
     up: async (queryInterface, Sequelize) => {
         await queryInterface.sequelize.transaction(async (transaction) => {
             await queryInterface.addColumn(table, 'templateId', {
-                type: Sequelize.INTEGER }, { transaction });
+                type: Sequelize.DOUBLE }, { transaction });
 
             await queryInterface.addIndex(table, ['templateId'], {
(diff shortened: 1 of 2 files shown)


jithine:
cc @parthasl

yoshwata:
What we should fix might be DOUBLE value of API.

InderH:
cc @yoshwata 

In datastore-sequelize index.js there is a function called getSequelizeTypeFromJoi that determines the type based on the joi data schema definitions. getSequelizeTypeFromJoi ignores the joi rule that the type should be an integer and instead just looks at the joi type which in this case happens to be 'number' and sets the sequalize data type for the field to be double. While templateId should be an integer (as per the data schema definitions), it is being set to double for this reason.

yoshwata:
Thanks @InderH. I confirmed on my local environment that the inconsistency between DB schema and API definition does not make any errors, so there is no problem for now.

PR #385, opened 2020-02-17
feat(1951): Update Schema for executor

pritamstyz4ever:
## Context

We need token to make api calls to the queue service and schema needs to allow passing to executor

## Objective

This PR allows token to be passed to executor to make api calls

## References

https://github.com/screwdriver-cd/screwdriver/issues/1951

## License


pritamstyz4ever:
> should the new fields be optional ? Do we envision cluster admins using an older version of executor-queue where `queue-service` is not used ?

Could be possible, also don't want to break anything when we merge.

PR #390, opened 2020-03-04
feat(1785): Add groupEventId [1]

tkyi:
## Context

Need a way to group similar events that we triggered by the same event or started manually.

## Objective

This PR adds a new field `groupEventId` to event model to track similar events. `parentEventId` cannot be used since it tracks the direct parent event id, not the master parent event id.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1785

## License

### DIFF
--- a/migrations/20200304-add-column_groupEventId_to_events.js
+++ b/migrations/20200304-add-column_groupEventId_to_events.js
@@ -3,16 +3,13 @@
 'use strict';
 
 const prefix = process.env.DATASTORE_SEQUELIZE_PREFIX || '';
-const table = `${prefix}builds`;
+const table = `${prefix}events`;
 
 module.exports = {
     up: async (queryInterface, Sequelize) => {
         await queryInterface.sequelize.transaction(async (transaction) => {
-            await queryInterface.addColumn(table, 'templateId', {
-                type: Sequelize.INTEGER }, { transaction });
-
-            await queryInterface.addIndex(table, ['templateId'], {
-                name: `${table}_template_id`, transaction });
+            await queryInterface.addColumn(table, 'groupEventId', {
+                type: Sequelize.DOUBLE }, { transaction });
         });
     }
 };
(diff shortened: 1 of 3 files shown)


models/event.js
```diff
@@ -18,7 +18,11 @@ const MODEL = {
         .example(123345),
     parentEventId: Joi
         .number().integer().positive()
-        .description('Identifier of the parent event')
+        .description('Identifier of the direct parent event')
+        .example(123344),
+    groupEventId: Joi
```
jithine: shouldn't this be optional in-case we merge this PR and rest of PRs are slow to merge ?

migrations/20200304-add-column_groupEventId_to_events.js
```diff
@@ -0,0 +1,18 @@
...
+
+'use strict';
+
+const prefix = process.env.DATASTORE_SEQUELIZE_PREFIX || '';
+const table = `${prefix}builds`;
+
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        await queryInterface.sequelize.transaction(async (transaction) => {
+            await queryInterface.addColumn(table, 'templateId', {
```
jithine: shouldn't this be `groupEventId`

models/event.js
klu909: this should be optional unless we are going to back fill older events, also we need db migration to add this column

tkyi: I don't think you need to mark fields in the model as optional.
You indicate "optional" fields below in the exports.

migrations/20200304-add-column_groupEventId_to_events.js
tkyi: renamed the file and github uploaded the wrong one
deleted the other one and the changes are reflected correctly now

models/event.js
jithine: where does that happen ?

tkyi: for example, these would be required: https://github.com/screwdriver-cd/data-schema/blob/master/models/event.js#L112
these would be optional: https://github.com/screwdriver-cd/data-schema/blob/master/models/event.js#L114-L115

it's done using `mutate`: https://github.com/screwdriver-cd/data-schema/blob/master/lib/mutate.js

PR #391, opened 2020-03-11
feat(1967): add query to datastore

InderH:
## Context

Need to add the ability to do raw queries to datastore so queries that are otherwise impossible to do through sequelize can be done, as such need to add a schema for this method to data schema.

## Objective

Add a schema for the datastore query method to data schema so method calls can be validated.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1967

## License

### DIFF
--- a/plugins/datastore.js
+++ b/plugins/datastore.js
@@ -41,6 +41,16 @@ const SCHEMA_SCAN = Joi.object().keys({
     timeKey: Joi.string(),
     aggregationField: Joi.string()
 });
+const SCHEMA_QUERY = Joi.object().keys({
+    table,
+    queries: Joi.array().items(
+        Joi.object().keys({
+            dbType: Joi.string(),
+            query: Joi.string()
+    })),
+    replacements: Joi.object(),
+    rawResponse: Joi.boolean()
+});
 
 module.exports = {
     /**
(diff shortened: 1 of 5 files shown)


plugins/datastore.js
```diff
@@ -41,6 +41,16 @@ const SCHEMA_SCAN = Joi.object().keys({
...
+const SCHEMA_QUERY = Joi.object().keys({
+    table,
+    queries: Joi.array().items(
+        Joi.object().keys({
+            dbType: Joi.string().required(),
+            query: Joi.string().required()
+        })).required(),
+    replacements: Joi.object(),
+    rawResponse: Joi.boolean()
+});
```
DekusDenial: missing `readOnly` from https://github.com/screwdriver-cd/models/pull/441/files#diff-43e1dd855c9ebdae88cbf0102c6e05b2R337

plugins/datastore.js
```diff
@@ -41,6 +41,16 @@ const SCHEMA_SCAN = Joi.object().keys({
...
     aggregationField: Joi.string()
 });
+const SCHEMA_QUERY = Joi.object().keys({
+    table,
+    queries: Joi.array().items(
+        Joi.object().keys({
+            dbType: Joi.string().required(),
+            query: Joi.string().required()
+        })).required(),
+    replacements: Joi.object(),
```
klu909: might want to be define what this object contains here

InderH: This object's keys would change based on the query, as such I don't think its possible to define what the object contains

PR #393, opened 2020-04-07
feat(1922): Add openPr property

klu909:
## Context

Adding a new method `openPr` in scm plugin

## Objective

Add validation for `openPr` method in scm plugin

## References

https://github.com/screwdriver-cd/scm-base/pull/77
https://github.com/screwdriver-cd/screwdriver/issues/1922

## License

### DIFF
--- a/plugins/scm.js
+++ b/plugins/scm.js
@@ -171,10 +171,12 @@ const GET_BRANCH_LIST = Joi.object().keys({
 const OPEN_PR = Joi.object().keys({
     scmUri,
     token,
-    files: Joi.array().items(Joi.objects().keys({
-        fileName: Joi.string().required(),
-        fileContent: Joi.string().required()
-    })).required(),
+    files: Joi.array().items(
+        Joi.object().keys({
+            fileName: Joi.string().required(),
+            fileContent: Joi.string().required()
+        })
+    ).min(1).required(),
     scmContext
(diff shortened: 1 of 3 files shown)


plugins/scm.js
```diff
@@ -168,6 +168,20 @@ const GET_BRANCH_LIST = Joi.object().keys({
...
+    files: Joi.array().items(
+        Joi.object().keys({
+            name: Joi.string().required(),
+            content: Joi.string().required()
+        })
+    ).min(1).required(),
+    scmContext,
+    title: Joi.string().required(),
+    message: Joi.string().required()
+}).required();
```
adong, line 183: I can understand the attributes of `OPEN_PR` are required, but why is `OPEN_PR` itself is `required()`? Would it be okay to use `optional()` instead?

adong: @tkyi might know better for `data-schema` changes.

adong: I've spoken with Tiffany offline, and confirmed it ought not to be breaking anything. Since this `OPEN_PR` will be used for checking params in
https://github.com/screwdriver-cd/screwdriver/pull/2075/files#diff-b2fdd2f3a4a83b3e213a4257e3ca15fdR57, and will not be stored in DB and cause other effects.

Approved.

PR #395, opened 2020-05-05
fix(1785): Add groupEventId index for events

tkyi:
## Context

Adding `groupEventId` as a field to index to events model will help with faster lookup.

## Objective

This PR adds `groupEventId` to events index.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1785

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #399, opened 2020-06-11
fix: Add index groupEventId to events

kkokufud:
## Context

The specific query of remote join is slowly without `groupEventId` index. Therefore we need to add `groupEventId` index to events. 

## Objective
Add `groupEventId` index to events. 

## References
https://github.com/screwdriver-cd/data-schema/pull/395

## License

### DIFF
--- a/migrations/20200611-add_index_groupEventId_to_events.js
+++ b/migrations/20200611-add_index_groupEventId_to_events.js
@@ -1,0 +1,16 @@
+/* eslint-disable new-cap */
+
+'use strict';
+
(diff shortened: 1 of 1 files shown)


kumada626:
If `DATASTORE_DDL_SYNC_ENABLED` is false, sequalize does not create this index automatically. And we expressly add index in migration file like [this](https://github.com/screwdriver-cd/data-schema/blob/913ddd116837b93c961dfae58e1b67d4f6ef31cd/migrations/20190919-initdb-events.js#L71-L90). So I think we should add this expressly by migration file.

jithine:
Missed that aspect. Thank you for catching this.

PR #401, opened 2020-06-16
fix: Add schema for jwt token config

pritamstyz4ever:
## Context

Jwt token config used for auth needs a defined schema which will validate any changes to the token structure

## Objective

This PR adds token config schema and additional properties to executor schema

## License

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -9,7 +9,17 @@ const regex = require('./regex');
 const template = require('./template');
 const workflowGraph = require('./workflowGraph');
 const parameters = require('./parameters');
+const tokenConfig = require('./tokenConfig');
 
 module.exports = {
-    annotations, base, command, commandFormat, job, regex, template, workflowGraph, parameters
+    annotations,
+    base,
+    command,
+    commandFormat,
+    job,
+    regex,
+    template,
+    workflowGraph,
+    parameters,
+    tokenConfig
 };
--- a/config/tokenConfig.js
+++ b/config/tokenConfig.js
@@ -1,0 +1,22 @@
+'use strict';
+
+const Joi = require('joi');
+const models = require('../models');
+
+const TOKEN_CONFIG = Joi.object({
+    username: [Joi.string().required(), Joi.ref('buildId')],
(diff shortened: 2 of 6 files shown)


config/tokenConfig.js
```diff
@@ -0,0 +1,22 @@
...
+
+const Joi = require('joi');
+const models = require('../models');
+
+const TOKEN_CONFIG = {
+    username: Joi.string(),
+    buildId: Joi.reach(models.build.base, 'id').required(),
+    jobId: Joi.reach(models.job.base, 'id').required(),
+    eventId: Joi.reach(models.event.base, 'id'),
+    isPR: Joi.boolean().required().default(false),
```
jithine, line 11: we can have default true, it's safer on auth levels. in-case there is a mistake  ?

plugins/executor.js
```diff
@@ -38,7 +38,9 @@ const SCHEMA_START = Joi.object().keys({
         .label('API URI'),
     token: Joi.string().required()
         .label('Build JWT'),
-    enqueueTime: Joi.date().iso()
+    enqueueTime: Joi.date().iso(),
+    isPR: Joi.boolean().optional().default(false),
```
jithine, line 42: I would say default `true` ? in case some caller forgets to set this, this will fall back on lower auth level

config/tokenConfig.js
pritamstyz4ever: ok, should this always be true, meaning all requests are PR's?

jithine: no caller should take care of passing it ?

jithine: does it have to have a `default` value ?

config/index.js
```diff
@@ -9,7 +9,17 @@ const regex = require('./regex');
...
+    annotations,
+    base,
+    command,
+    commandFormat,
+    job,
+    regex,
+    template,
+    workflowGraph,
+    parameters,
+    tokenConfig
```
tkyi, line 24: I think these should be alphabetical order, but it's not a big deal

PR #402, opened 2020-06-18
fix: Add parentEventId index

sakka2:
## Context
The specific query of remote join is slowly without parentEventId index. 
It was not enough to add the index groupEventId.
Therefore we need to add parentEventId index to events.

## Objective
Add parentEventId index to events.

## References
#399 

## License


kumada626:
Also you need to add index in this file.
https://github.com/screwdriver-cd/data-schema/blob/0d1094bb2060b3f8071ab94239683518c656aa6b/models/event.js#L165-L166

kkisic:
> Also you need to add index in this file.
> https://github.com/screwdriver-cd/data-schema/blob/0d1094bb2060b3f8071ab94239683518c656aa6b/models/event.js#L165-L166

Thx. I fixed it.

PR #404, opened 2020-07-17
feat(1079): Validation schemas for deploy key pipeline [1]

supra08:
## Context

Fixes screwdriver-cd/screwdriver#1079

## Objective

This PR adds relevant fields in the schemas for handling autoDeployKeyGeneration flag.

## References

screwdriver-cd/screwdriver#1079

## License

### DIFF
--- a/test/data/auth.contexts.yaml
+++ b/test/data/auth.contexts.yaml
@@ -2,3 +2,4 @@
 -
   context: this is a context,
   displayName: this is a display name
+  autoDeployKeyGeneration: true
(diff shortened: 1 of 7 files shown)


test/data/auth.contexts.yaml
```diff
@@ -2,3 +2,4 @@
 -
   context: this is a context,
   displayName: this is a display name
+  autoDeployKeyGeneration: true
```
tkyi, line 5: Minor comment, but just wondering why there's `autoDeployKeyGeneration` vs `autoKeysGeneration`?
Would it be better to make them the same?

supra08: So actually these keys actually carry flags from 2 different sources. `autoDeployKeyGeneration` from the API config and `autoKeysGeneration` from the UI after the user has confirmed. So I kept them different in order to differentiate them. But I guess the UI flag can be renamed to something better.

adong: Yeah, I had similar confusion as well. https://github.com/screwdriver-cd/ui/pull/568#discussion_r465889020

PR #406, opened 2020-07-31
feat(1958): Upgrade to @hapi/hapi v19.x.x. BREAKING CHANGE: pulling in joi v17

pritamstyz4ever:
## Context

Screwdriver hapi.js dependencies are outdated and not supported anymore.

## Objective

This PR majorly upgrades @hapi/joi version and fixes all schema related changes to be compatible with latest @hapi js dependencies.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1958

## License

### DIFF
--- a/api/auth.js
+++ b/api/auth.js
@@ -17,7 +17,8 @@ const SCHEMA_CRUMB = Joi.object().keys({
 const SCHEMA_CONTEXTS = Joi.array().items(
     Joi.object().keys({
         context: Joi.string().label('Context name'),
-        displayName: Joi.string().label('Display name')
+        displayName: Joi.string().label('Display name'),
+        autoDeployKeyGeneration: Joi.boolean().label('Auto deploy key generation flag')
     }).label('Context Object')
 ).label('Array of Contexts');
 
--- a/api/auth.js
+++ b/api/auth.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const Joi = require('@hapi/joi');
+const Joi = require('joi');
 
 const SCHEMA_KEY = Joi.object().keys({
     key: Joi.string().label('Public Key for Verifying JWTs')
--- a/api/auth.js
+++ b/api/auth.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const Joi = require('joi');
+const Joi = require('@hapi/joi');
 
 const SCHEMA_KEY = Joi.object().keys({
     key: Joi.string().label('Public Key for Verifying JWTs')
--- a/api/commandValidator.js
+++ b/api/commandValidator.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const Joi = require('@hapi/joi');
+const Joi = require('joi');
 
 // essentially a Joi.ValidationError.details object
 const COMMAND_ERROR = Joi.object()
--- a/api/commandValidator.js
+++ b/api/commandValidator.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const Joi = require('joi');
+const Joi = require('@hapi/joi');
 
 // essentially a Joi.ValidationError.details object
 const COMMAND_ERROR = Joi.object()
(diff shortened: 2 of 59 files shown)


api/auth.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const Joi = require('joi');
+const Joi = require('@hapi/joi');
```
adong, line 3: Just want to point out.
image


Ref: https://github.com/sideway/joi/issues/2411

jithine: from what I read going to v19 is the first step, so this change will be required

jithine: but you are right, we can omit `hapi@` module and just use `joi` with v17

api/commandValidator.js
```diff
@@ -1,32 +1,34 @@
 'use strict';
 
-const Joi = require('joi');
+const Joi = require('@hapi/joi');
```
jithine, line 3: this can me reverted to `require('joi');` to avoid future changes.

pritamstyz4ever: I think we should wait until this is resolved.
https://github.com/sideway/joi/issues/2411

jithine: so what he's saying is  that ```The npm joi module is being resurrected and the current v17.1.1 has been published there (identical to the version available on @hapi/joi).```
current `@hapi/joi` & `joi` in npm are same

config/job.js
```diff
@@ -3,25 +3,24 @@
...
-// ref. https://github.com/hapijs/joi/blob/v13.3.0/API.md#extendextension
+// ref. https://github.com/sideway/joi/blob/v17.1.1/API.md#extendextensions
 const sdJoi = Joi.extend(joi => ({
+    type: 'string',
     base: joi.string(),
-    name: 'string',
-    language: {
-        branchFilter: 'invalid trigger format'
+    messages: {
+        'string.branchFilter': '"{{#label}}" has invalid trigger format {{#q}}'
```
tkyi, line 15:
```suggestion
        'string.branchFilter': '{{#label}} has invalid trigger format {{#q}}'
```

core/scm.js
```diff
@@ -164,14 +166,18 @@ const SCHEMA_HOOK = Joi.object().keys({
 
     ref: Joi.string()
         .when('action',
-            { is: ['release', 'tag'], then: Joi.required(), otherwise: Joi.allow('').optional() })
+            {
+                is: Joi.valid('release', 'tag'),
+                then: Joi.required(),
+                otherwise: Joi.optional()
```
tkyi, line 172: missing `allow('')` for otherwise case?

models/banner.js
```diff
@@ -33,12 +33,7 @@ const MODEL = {
         .description('Username of user creating the banner')
         .example('batman123'),
 
-    type: Joi
-        .string().valid([
-            'info',
-            'warn'
-        ])
-        .max(32)
+    type: Joi.string().valid('info', 'warn')
```
tkyi, line 36: might still need max length for https://github.com/screwdriver-cd/screwdriver/issues/325?
actually nvm looks like this PR should check: https://github.com/screwdriver-cd/data-schema/pull/280/files
```suggestion
    type: Joi.string().valid('info', 'warn').max(32)
```

pritamstyz4ever: @tkyi The .max() function is not allowed for .valid(). It has two values that should validate only info and warn.

config/template.js
```diff
@@ -56,27 +57,21 @@ const TEMPLATE_MAINTAINER = Joi
...
-        config: Job.job,
+        name: TEMPLATE_NAME.required(),
+        version: TEMPLATE_VERSION.required(),
+        description: TEMPLATE_DESCRIPTION.required(),
+        maintainer: TEMPLATE_MAINTAINER.required(),
+        config: Job.job.required(),
         images: TEMPLATE_IMAGES
-    })
-    .requiredKeys('name', 'version', 'description', 'maintainer',
-        'config', 'config.image', 'config.steps');
```
tkyi, line 79: is there any way to account for the requires `config.image` and `config.steps`?

pritamstyz4ever: Added the check for `config.image` and `config.steps`. 
Although, there is hardly any documentation for requiredKeys alternative in hapi 17, we may need to update some base schemas in some cases.

PR #412, opened 2020-09-17
feat(2199): Add pipeline and user settings [1]

tkyi:
## Context

Users might want to configure certain things in the UI, such as job name length or which jobs to watch as a part of a downtime metric.

## Objective

This PR adds settings to user and pipeline models.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2199, https://github.com/screwdriver-cd/ui/pull/596

## License


models/user.js
```diff
@@ -23,7 +23,12 @@ const MODEL = {
         .string()
         .max(128)
         .description('The SCM to which the user belongs')
-        .example('github:github.com')
+        .example('github:github.com'),
+
+    settings: Joi
+        .object()
+        .description('User settings')
+        .example({ display: { jobNameLength: 20 } })
```
adong, line 31: I was thinking for user preference schema like: 

```
settings: {
   pipelines: [{
	 id: 1, // or pipelinId, either way
   	 jobNameLength: 20
   }, {
	 id: 2,
   	 jobNameLength: 30
   }]
   x: 'y', // this can be used later as global settings, rather than pipelines specifics
}
```

tkyi: I guess it depends if you think pipeline settings should be personalized per user or per pipeline
like repo settings in Github are tied to the repo, not the user

tkyi:
Note: need to add migration files

adong (approved):

config/settings.js
```diff
@@ -0,0 +1,30 @@
...
+    .default(20)
+    .description('Job name length to display on workflowGraph')
+    .example(20);
+const SCHEMA_METRICS_DOWNTIME_JOBS = Joi.array().items(
+    Job.jobName
+).description('Jobs to watch for downtime');
+
+const SCHEMA_PIPELINE_SETTINGS = Joi.object()
+    .keys({
+        metricsDowntimeJobs: SCHEMA_METRICS_DOWNTIME_JOBS
```
jithine, line 18: should this be a generic object to store any sort of key-value setting ?

tkyi: sure
I can make it generic if you want to have no validation on `settings`

jithine: ok, dev needs to be aware of this validation requirement if adding a new setting. I guess this is fine, even if we relax this later, there shouldn't be any schema changes right ?

jithine (approved):

config/settings.js
tkyi: no I don't think so, since we wouldn't be adding any columns

PR #417, opened 2020-10-20
fix(2199): Downtime jobs should be array of jobIds [1]

tkyi:
## Context

Would be better for UI to pass in array of job IDs rather than have the API look them up.

## Objective

This PR changes `downtimeJobs` to array of `jobIds` vs `jobNames`; also adds `downtimeStatuses`.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2199

## License

### DIFF
--- a/config/settings.js
+++ b/config/settings.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const Joi = require('joi');
-const Job = require('./job');
 
 const SCHEMA_DISPLAY_JOB_NAME_LENGTH = Joi.number().integer()
     .min(20)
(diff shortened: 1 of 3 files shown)


pritamstyz4ever (approved):

jithine (approved):
WIll jobIds change when job jobs are added or exo

tkyi:
@jithine 
By default, the API will look through all jobs, and nothing will need to be done to keep that up to date.

If the user specifies certain `downtimeJobs`, the UI will need to update the `jobIds` passed in.

PR #431, opened 2021-01-27
feat(2319): [1] Add order field to job config, warnMessages for template validator

tkyi:
## Context

Would be nice if template owners could pick and choose which steps from a template should be run in their own.

## Objective

This PR adds support for the `order` keyword in a job; also adds support for warnMessages for template validator.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2319

## License

### DIFF
--- a/api/templateValidator.js
+++ b/api/templateValidator.js
@@ -28,7 +28,8 @@ const SCHEMA_OUTPUT = Joi.object()
         // since a template could be parseable but invalid, the contents are unpredictable
         template: Joi.object()
             .label('The end-result of parsing the given template')
-            .required()
+            .required(),
+        warnMessages: Joi.array().optional()
     })
     .label('Template validation output');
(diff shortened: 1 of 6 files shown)


klu909 (approved):

api/templateValidator.js
```diff
@@ -28,7 +28,8 @@ const SCHEMA_OUTPUT = Joi.object()
         // since a template could be parseable but invalid, the contents are unpredictable
         template: Joi.object()
             .label('The end-result of parsing the given template')
-            .required()
+            .required(),
+        warnMessages: Joi.array().optional()
```
jithine, line 32: @tkyi should we keep a generic parseErrors object ? We can eventually use it to store config parse errors.  https://github.com/screwdriver-cd/screwdriver/issues/2326

PR #437, opened 2021-02-12
feat(2331): Expand step object for job validation [1]

tkyi:
## Context

It would be nice if template owners could indicate which steps cannot be overwritten in their templates. In order to allow for this change (and other step expansions), it would be nice if steps can be expanded to more than just a string or single key-value object.

## Objective

This PR adds the option to expand step object from previous formats `'npm publish'`/`{ deploy: npm publish }` to allow for more complex objects `{ deploy: { command: 'npm publish', locked: true } }`.
Also adds `locked` field to step.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2331

## License


config/job.js
```diff
@@ -77,34 +77,61 @@ const SCHEMA_ENVIRONMENT = Joi.object()
...
     });
 const SCHEMA_JOBNAME = Joi.string().max(100).regex(Regex.JOB_NAME);
+// Step can be in the following formats:
+// npm install
+// { init: npm install }
+// { init: { command: npm install } }
 const SCHEMA_STEP_STRING = Joi.string();
+const SCHEMA_STEP_SUBOBJECT_BASE = Joi.object()
+    .keys({
+        command: SCHEMA_STEP_STRING.required()
```
jithine, line 87: does job step need to have locked to determine that it's not overridable ? Or will we be looking into template step  config for this info ?

migrations/20210212-add-column_locked_to_steps.js
```diff
@@ -0,0 +1,15 @@
...
+
+'use strict';
+
+const prefix = process.env.DATASTORE_SEQUELIZE_PREFIX || '';
+const table = `${prefix}steps`;
+
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        await queryInterface.sequelize.transaction(async (transaction) => {
+            await queryInterface.addColumn(table, 'locked', {
```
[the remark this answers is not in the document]
tkyi: I think that's better for validation and quicker lookup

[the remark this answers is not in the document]
tkyi: but for template key probably don't need that actually

config/job.js
tkyi: this will only be used for validation

but job step doesn't need `locked`
removed

config/job.js
```diff
@@ -77,34 +77,61 @@ const SCHEMA_ENVIRONMENT = Joi.object()
...
-    .items(SCHEMA_STEP_STRING.regex(Regex.STEP_NAME))
-    .min(0);
-const SCHEMA_SETTINGS = Joi.object().optional();
 const SCHEMA_STEP = Joi.alternatives().try(SCHEMA_STEP_STRING, SCHEMA_STEP_OBJECT);
 const SCHEMA_STEPS = Joi.array().items(SCHEMA_STEP).min(1);
-const SCHEMA_STEPS_NO_DUPS = Joi.array().items(SCHEMA_STEP).min(1).unique((a, b) => {
+// Template steps
+const SCHEMA_TEMPLATE_STEP_SUBOBJECT = SCHEMA_STEP_SUBOBJECT_BASE
+    .keys({
+        locked: Joi.boolean()
```
jithine, line 106: should this be optional ? should be allow users to define the step object format without any attributes ?

config/job.js
```diff
@@ -77,34 +77,61 @@ const SCHEMA_ENVIRONMENT = Joi.object()
...
+    });
+const SCHEMA_TEMPLATE_STEP_COMMAND = Joi.alternatives()
+    .try(SCHEMA_STEP_STRING, SCHEMA_TEMPLATE_STEP_SUBOBJECT);
+const SCHEMA_TEMPLATE_STEP_OBJECT = Joi.object()
+    // Steps can only be named with A-Z,a-z,0-9,-,_
+    // Steps only contain strings (the command to execute)
+    .pattern(Joi.string().regex(/^[\w-]+$/), SCHEMA_TEMPLATE_STEP_COMMAND)
```
jithine, line 113: is this regex same as `Regex.STEP_NAME` ?

config/job.js
tkyi: I think it's okay
the template-validator/config-parser can convert the non-attribute step objects to the simplified version

jithine (approved):

PR #439, opened 2021-02-22
fix: update user settings

adong:
## Context

add to allow arbitrary keys, and add `showPRJobs` for individual pipeline settings in user settings.

## Objective

* enforce pipeline settings
* update test

image

image

## References

[Joi Tester](https://framp.me/joi-tester/)
## License

### DIFF
--- a/test/data/user.get.yaml
+++ b/test/data/user.get.yaml
@@ -3,4 +3,10 @@ id: 123
 username: batman
 scmContext: github:github.com
 settings:
-    displayJobNameLength: 20
+    1:
+        displayJobNameLength: 20
+    11:
+        displayJobNameLength: 50
+        showPRJobs: false
+    hello: world
+    this: is-allowed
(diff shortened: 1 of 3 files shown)


test/data/user.get.yaml
```diff
@@ -3,4 +3,10 @@ id: 123
 username: batman
 scmContext: github:github.com
 settings:
-    displayJobNameLength: 20
+    1:
```
tkyi, line 6: what is this number referring to?

pritamstyz4ever (approved):

test/data/user.get.yaml
adong: 1 means pipeline number. It would be like:
```
1: {
  displayJobNameLength: 10,
  showPRJobs: true
},
11: {
  displayJobNameLength: 10,
  showPRJobs: true
},
some_arbitary_vars: 'xyz'
```

PR #440, opened 2021-02-22
fix(2331): Pass locked job steps [2.5]

tkyi:
## Context

Would be nice if the UI could show locked steps in a job when template is used.

## Objective

This PR allows more data to be sent from config parser through commands so locked step can be shown in the UI.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2331

## License

### DIFF
--- a/api/validator.js
+++ b/api/validator.js
@@ -13,7 +13,7 @@ const SCHEMA_JOB_COMMAND = Joi.object()
         name: Joi.string(),
         command: Joi.string()
     })
-    .unknown(false)
+    .unknown(true) // allow other fields
     .label('Named command to execute');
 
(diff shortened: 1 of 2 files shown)


pritamstyz4ever (approved):

api/validator.js
```diff
@@ -13,7 +13,7 @@ const SCHEMA_JOB_COMMAND = Joi.object()
         name: Joi.string(),
         command: Joi.string()
     })
-    .unknown(false)
```
jithine, line 16: should we just add `locked` as optional here, instead of allowing `unknown` ?

tkyi: added

jithine (approved):

PR #450, opened 2021-06-14
feat: Add github & gitlab PR Status validations

ombharatiya:
## Context

Add Github & Gitlab PR Status validation for PENDING state. 

## Objective

Add both SCM specific status lists for the validation check. Adding the Pending State in Meta goes through the validation checks before reflecting on the PR Build details.

https://docs.screwdriver.cd/user-guide/metadata#additional-pull-request-checks need to support PENDING status

## References

Jira:
https://github.com/screwdriver-cd/screwdriver/issues/2449

More related PRs:
screwdriver-cd/scm-github#189

screwdriver-cd/models#496

https://github.com/screwdriver-cd/models/pull/506

https://github.com/screwdriver-cd/scm-github/pull/190

https://github.com/screwdriver-cd/scm-gitlab/pull/43

https://github.com/screwdriver-cd/data-schema/pull/450

## License


plugins/scm.js
```diff
@@ -21,6 +20,20 @@ const token = models.user.base.extract('token').required();
 const type = core.scm.hook.extract('type').required();
 const username = models.user.base.extract('username').required();
 
+const githubStatus = [
+    'PENDING',
```
jithine, line 24: does letter case matter here ?

plugins/scm.js
```diff
@@ -107,7 +120,7 @@ const UPDATE_COMMIT_STATUS = Joi.object().keys({
     scmUri,
     token,
     sha,
-    buildStatus,
+    buildStatus: Joi.string().valid(...githubStatus, ...gitlabStatus),
```
pritamstyz4ever, line 123: is this supposed to be a mandatory field?

plugins/scm.js
ombharatiya: Capital Letter is working as I have verified that. But the `UPDATE_COMMIT_STATUS` object is being used to validate status of builds as well where it is conflicting the purpose. I'm checking that.

plugins/scm.js
ombharatiya: Yes, I think there is no scenario where we can update a commit without its status.

plugins/scm.js
ombharatiya: It is being handled in scm github & gitlab individually.

At other places capital letters are being used.

plugins/scm.js
ombharatiya: @jithine do you want to add anything here?

pritamstyz4ever: then it should have a  .required() I believe

plugins/scm.js
```diff
@@ -21,6 +20,12 @@ const token = models.user.base.extract('token').required();
 const type = core.scm.hook.extract('type').required();
 const username = models.user.base.extract('username').required();
 
+const githubStatus = [
```
tkyi, line 23: is this is specific to github, what happens for gitlab?

jithine: That variable should be renaned https://github.com/screwdriver-cd/data-schema/pull/450#pullrequestreview-688969233

Idea is that API/Model should deal with only success/pending/failure as supported SCM status checks values. 
meta staus checks & build status should align with these values before being sent to SCM methods (updateCommitStatus). Then it's up to the individual SCM's to map it to values which pertains to it's individual SCM API values.

plugins/scm.js
```diff
@@ -120,7 +112,7 @@ const UPDATE_COMMIT_STATUS = Joi.object().keys({
     scmUri,
     token,
     sha,
-    buildStatus: Joi.string().valid(...githubStatus, ...gitlabStatus),
+    buildStatus: Joi.string().valid(...githubStatus),
```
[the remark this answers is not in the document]
ombharatiya: Renamed it to `scmStatus`. Thank you.

plugins/scm.js
ombharatiya: Yes, renamed to `scmStatus`.

We've already mapped 'failed' state with FAILURE in Gitlab's methods. 

Please refer to this:
https://github.com/screwdriver-cd/scm-gitlab/pull/43

plugins/scm.js
ombharatiya: Added the required() validation. Thank you.

plugins/scm.js
tkyi: So we're keeping track of this map in data-schema, models, and each scm-plugin;
but the one in data-schema is specific to github?

jithine: I think it's better to have proper separation of concerns

pritamstyz4ever (commented):
Please add test

plugins/scm.js
pritamstyz4ever: I feel schema should be the place where we define things, but it should not be made specific to Github. Also keeping track in multiple places has a high potential of issues. We already are using schema in all places and should refer to status schema from here, instead of const strings everywhere

lib/scmCommons.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+/**
+ * Scm spcific common objects
```
tkyi, line 4:
```suggestion
 * Scm specific common objects
```

tkyi (changes requested):
Can you move the scm constants to either plugins/scm or core/scm?
Seems to be a couple files where scm stuff is kept and want to keep it to two

lib/scmCommons.js
```diff
@@ -0,0 +1,23 @@
...
+ * @type {Object}
+ */
+module.exports = {
```
tkyi, line 7: need to delete this file

ombharatiya: Deleted. Thanks.

tkyi (approved):

PR #459, opened 2021-09-01
fix: Add description to autoKeysGeneration

tkyi:
## Context

It's confusing what certain fields are for in the API.
Screen Shot 2021-09-01 at 3 29 14 PM

## Objective

This PR adds a description to the field autoKeysGeneration.

## References

N/A
## License

### DIFF
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -19,7 +19,9 @@ const CREATE_MODEL = {
         .required(),
 
     rootDir: Scm.rootDir,
-    autoKeysGeneration: Joi.boolean().optional()
+    autoKeysGeneration: Joi.boolean()
+        .optional()
+        .description('Flag to generate and add the public deploy key for a specific scm')
 };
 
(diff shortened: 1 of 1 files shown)


models/pipeline.js
```diff
@@ -19,7 +19,7 @@ const CREATE_MODEL = {
         .required(),
 
     rootDir: Scm.rootDir,
-    autoKeysGeneration: Joi.boolean().optional()
+    autoKeysGeneration: Joi.boolean().optional().description('Flag to generate and add the public deploy key for a specific scm (used for private repos)')
```
jithine, line 22: strictly not only for private repos. Any action which needs auth, eg: git clone for non-public SCM (Git hub enterprise vs github.com) needs deploy key.
Also is needed when build wants to commit back to SCM (for public repos)

jithine (approved):

adong (approved):

PR #462, opened 2021-10-04
fix: Bump up max char length for triggers

tkyi:
## Context

The [max job name length](https://github.com/screwdriver-cd/data-schema/blob/master/config/job.js#L81) is 100 chars, but the [max trigger name length](https://github.com/screwdriver-cd/data-schema/blob/master/models/trigger.js#L12-L31) is only 64 chars (including the additional `~sd@` prefix).

## Objective

This PR bumps up the max trigger name length to 128. 

## References
N/A

## License

### DIFF
--- a/migrations/20211004-update_trigger_length-triggers.js
+++ b/migrations/20211004-update_trigger_length-triggers.js
@@ -1,0 +1,17 @@
+/* eslint-disable new-cap */
+
+'use strict';
+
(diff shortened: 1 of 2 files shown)


jithine (commented):
Would need a migration file to go with this in order to update https://github.com/screwdriver-cd/data-schema/blob/master/migrations/20190919-initdb-triggers.js#L20-L25

jithine (approved):

PR #464, opened 2021-11-02
fix: Allow other fields for executor params

tkyi:
## Context

Should not fail if extra params are passed into executor.

## Objective

This PR adds allowUnknown to executor params.

## References
N/A

## License

### DIFF
--- a/plugins/executor.js
+++ b/plugins/executor.js
@@ -52,6 +52,7 @@ const buildSchemaObj = {
 };
 const SCHEMA_START = Joi.object()
     .keys(buildSchemaObj)
+    .unknown(true) // allow other fields
     .required();
 const SCHEMA_STOP = Joi.object()
     .keys({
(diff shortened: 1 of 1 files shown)


jithine (approved):

tk3fftk:
@tkyi @jithine 
Could you please tell us about this fix's motivation? :eyes:

jithine:
@tk3fftk this was to address a backward incompatibility when the changes were being rolled out. We had an outage when we pulled in new API because executor changes were not ready yet.

tk3fftk:
@jithine thanks. you mean it's temporary fix, right?

jithine:
not exactly temporary, by allowing unknown we give flexibility during active development  to make changes. Missing params should definitely throw error,  having `unknown` enables us to pass extra params which might not be prime time ready

PR #477, opened 2022-02-04
fix(2640): add pipelineId to unique key constraint

jithine:
## Context

When a single PR generates multiple events across different pipelines for same repository ([source dir](https://docs.screwdriver.cd/user-guide/configuration/sourceDirectory)), some of the event creation can fail, since `createTime` & `sha` is a unique constraint. Multiple events can get same `createTime` since it's [generated by API ](https://github.com/screwdriver-cd/models/blob/master/lib/eventFactory.js#L802)

## Objective

Add pipeline Id to unique Key constraint so that event creation for multiple pipelines caused by same PR  (due to Source Dir feature) can work. 

## References

https://github.com/screwdriver-cd/screwdriver/issues/2640

## License

### DIFF
--- a/migrations/20220203-uniquekey-update-events.js
+++ b/migrations/20220203-uniquekey-update-events.js
@@ -1,0 +1,21 @@
+/* eslint-disable new-cap */
+
+'use strict';
+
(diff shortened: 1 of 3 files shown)


sagar1312 (approved):
👍

PR #481, opened 2022-03-02
fix: modify migration files to current sequelize params

tk3fftk:
## Context

migration for new screwdriver cluster is not working due to sequelize's params change.

the error is like below:

> ERROR: Fields must be specified through options.fields

## Objective

modify old migration files to current sequelize params.
it passed migration for an empty database (using MySQL 5.7.33).

## References

- [QueryInterface | Sequelize](https://sequelize.org/master/class/src/dialects/abstract/query-interface.js~QueryInterface.html#instance-method-addConstraint)

## License

### DIFF
--- a/migrations/20190919-initdb-banners.js
+++ b/migrations/20190919-initdb-banners.js
@@ -36,7 +36,8 @@ module.exports = {
                 { transaction }
             );
 
-            await queryInterface.addConstraint(table, ['message', 'createTime', 'type'], {
+            await queryInterface.addConstraint(table, {
+                fields: ['message', 'createTime', 'type'],
                 name: `${table}_message_createTime_type_key`,
                 type: 'unique',
                 transaction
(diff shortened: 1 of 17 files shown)


migrations/20190919-initdb-banners.js
```diff
@@ -36,7 +36,8 @@ module.exports = {
                 { transaction }
             );
 
-            await queryInterface.addConstraint(table, ['message', 'createTime', 'type'], {
+            await queryInterface.addConstraint(table, {
+                fields: ['message', 'createTime', 'type'],
```
yuichi10, line 40: [nits]
I feel that the field key should be placed under the name key because it fits with the other settings.

tk3fftk: fixed 🙆‍♂️

kumada626 (approved):

tkyi (approved):

PR #485, opened 2022-03-15
fix(2676): Allow provider path to external repo

tkyi:
## Context

Users might want to point their provider to an external repo rather than copy paste a bunch of repetitive code.

## Objective

This PR allows for a checkout url type string as the provider in a screwdriver.yaml.

before:

```
provider:
    name: aws
    region: us-west-2
    accountId: 111111111111
    role: arn:aws:iam::111111111111:role/role
    executor: eks
    clusterName: sd-build-eks
```

after:
`provider: git@github.com:tkyi/testing.git#test:provider.yaml`

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2676

## License

### DIFF
--- a/plugins/executor.js
+++ b/plugins/executor.js
@@ -3,6 +3,7 @@
 const Joi = require('joi');
 const Annotations = require('../config/annotations');
 const Job = require('../config/job');
+const Provider = require('../config/provider');
 const models = require('../models');
 const buildId = models.build.base.extract('id').required();
 const eventId = models.event.base.extract('id');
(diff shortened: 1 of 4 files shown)


adong (approved):

plugins/executor.js
```diff
@@ -26,7 +27,7 @@ const buildSchemaObj = {
     jobName,
     jobState,
     jobArchived,
-    provider: Job.provider,
+    provider: Provider.provider.optional(),
```
adong, line 30: Hmm, it makes `provider` optional, is it supposed to be mandatory? 

image

tkyi: Provider field can be optional 
The params under provider are not optional if provider is configured

jithine (approved):

PR #486, opened 2022-03-22
feat(1712): Relaxing blockedBy for same job

klu909:
## Context

adding the option for job blocking behavior
## Objective

adding the option for job blocking behavior
## References

https://github.com/screwdriver-cd/screwdriver/issues/1712
## License


config/job.js
```diff
@@ -163,6 +163,12 @@ const SCHEMA_BLOCKEDBY_VALUE = Joi.alternatives().try(
 );
 const SCHEMA_BLOCKEDBY = Joi.alternatives().try(Joi.array().items(SCHEMA_BLOCKEDBY_VALUE), SCHEMA_BLOCKEDBY_VALUE);
 const SCHEMA_FREEZEWINDOWS = Joi.alternatives().try(Joi.array().items(SCHEMA_CRON_EXPRESSION), SCHEMA_CRON_EXPRESSION);
+const SCHEMA_BLOCKEDBYSAMEJOB = Joi.boolean().optional();
+const SCHEMA_BLOCKEDBYSAMEJOBWAITTIME = Joi.number()
+    .integer()
+    .positive()
+    .optional()
+    .allow(null);
```
tkyi, line 171: should we add defaults here?

klu909: default is handled in queue-service, but yeah we can add it here too

config/job.js
```diff
@@ -180,6 +186,8 @@ const SCHEMA_JOB = Joi.object()
         description: SCHEMA_DESCRIPTION,
         environment: SCHEMA_ENVIRONMENT,
         freezeWindows: SCHEMA_FREEZEWINDOWS,
+        blockedBySameJob: SCHEMA_BLOCKEDBYSAMEJOB,
+        blockedBySameJobWaitTime: SCHEMA_BLOCKEDBYSAMEJOBWAITTIME,
```
tkyi, line 190: it kind of seems weird to me that it's a job-level field vs annotation, but I guess `blockedBy` is also at this level

I think I would prefer these fields as annotations

klu909: Reasoning is that you probably wouldn't want this on every job

tkyi: could be job-level annotation, right?

tkyi (approved):

PR #491, opened 2022-06-13
feat(2669): Add stage model [1]

tkyi:
## Context

Users might want to group jobs together visually in the UI into stages.

## Objective

This PR adds stages table to the DB.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2669

## License

### DIFF
--- a/api/validator.js
+++ b/api/validator.js
@@ -39,6 +39,7 @@ const SCHEMA_JOB_PERMUTATION = Joi.object()
         secrets: Job.secrets,
         settings: Job.settings,
         sourcePaths: Job.sourcePaths,
+        stages: Base.stages,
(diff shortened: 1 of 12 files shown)


adong (approved):

api/validator.js
```diff
@@ -39,6 +39,7 @@ const SCHEMA_JOB_PERMUTATION = Joi.object()
         secrets: Job.secrets,
         settings: Job.settings,
         sourcePaths: Job.sourcePaths,
+        stages: Base.stages,
```
sagar1312, line 42: Job cannot be part of multiple stages.
This schema needs to be fixed so that a job points to a single stage.

PR #494, opened 2022-06-28
[skip ci] fix: index prefix for 20220607-initdb-stages.js

DekusDenial:


adong (approved):

migrations/20220607-initdb-stages.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
             });
 
             await queryInterface.addIndex(table, {
-                name: `${prefix}_pipeline_id`,
+                name: `${table}_pipeline_id`,
```
sagar1312, line 50: This updated script may be a noop on DB where the migration is already executed.
Should we create a new migration script to correct (drop old and create new) the index?

DekusDenial: These are just the names so really no impact at all. Given how fast we fixed this and stages is not completely built out yet and offered to use, this can be fixed by undoing last migration and redo it again.

PR #499, opened 2022-07-28
fix(2649): Remove unused column 'parameters' from 'templates' table

sagar1312:
## Context

`parameters` column in `templates` table is not being used. `parameters` is embedded into the config which is stored under the column `config`.

## Objective

This PR drops the column `parameters` from the `templates` table.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2649

## License

### DIFF
--- a/config/template.js
+++ b/config/template.js
@@ -3,7 +3,6 @@
 const Joi = require('joi');
 const Job = require('./job');
 const Regex = require('./regex');
-const Parameters = require('./parameters');
 
 const TEMPLATE_NAMESPACE = Joi.string()
     .regex(Regex.TEMPLATE_NAMESPACE)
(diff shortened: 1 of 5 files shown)


config/migrationsConfig.js
```diff
@@ -3,7 +3,8 @@
 module.exports = {
     development: {
         url: process.env.DEV_DATASTORE_SEQUELIZE_URL,
-        logging: true
+        logging: true,
+        dialect: 'mysql'
```
adong, line 7: Should we specify dialect here? If we hardcode to mysql, would there be another file for postgres?

sagar1312: Didn't intend to check-in this file. Fixed it.

tkyi (changes requested):
Do you need to remove `parameters` from here too: https://github.com/screwdriver-cd/data-schema/blob/master/models/template.js#L36

config/template.js
```diff
@@ -74,8 +71,7 @@ const SCHEMA_TEMPLATE = Joi.object().keys({
         .required()
         .or('image', 'template')
         .or('steps', 'template'),
-    images: TEMPLATE_IMAGES,
-    parameters: SCHEMA_JOB_PARAMETERS
```
tkyi, line 78: can users not put parameters in template?

sagar1312: yes, they can. It goes under `config` similar to `environment`, `settings` and other attributes that are allowed on the job.

tkyi (approved):

PR #501, opened 2022-08-08
fix: allow aliasname empty

banusha2:
## Context

User cannot rename the pipeline back to empty string, on changing AliasName to an empty string it gives `400`.

## Objective
Users can rename the pipeline alias back to an empty string.

## References
https://github.com/screwdriver-cd/screwdriver/issues/2744

## License

### DIFF
--- a/config/settings.js
+++ b/config/settings.js
@@ -2,6 +2,11 @@
 
 const Joi = require('joi');
 
+const SCHEMA_TIMESTAMP_FORMAT = Joi.string()
+    .valid('UTC', 'LOCAL_TIMEZONE', 'HUMAN_READABLE')
+    .optional()
+    .default('HUMAN_READABLE')
+    .description('User preferred timestamp');
 const SCHEMA_DISPLAY_JOB_NAME_LENGTH = Joi.number()
     .integer()
     .min(20)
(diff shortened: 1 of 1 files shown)


config/settings.js
```diff
@@ -38,6 +38,7 @@ const SCHEMA_PIPELINE_SETTINGS = Joi.object()
         showEventTriggers: Joi.boolean().optional(),
         filterEventsForNoBuilds: Joi.boolean().optional(),
         aliasName: Joi.string()
+            .allow('')
```
pritamstyz4ever, line 41: does it need to be empty or optional settings?

banusha2:
as per requirement it should be empty.

banusha2:
it's an optional field but later users want to rename it to an empty string
they should be able to do .

pritamstyz4ever:
> it's an optional field but later users want to rename it to an empty string they should be able to do . On Mon, Aug 8, 2022 at 1:27 PM Anusha Bollampally < ***@***.***> wrote:
> […](#)
> as per requirement it should be empty. 

Please add `.optional()` if it is an optional field

banusha2:
Please check .optional() exist already

config/settings.js
tkyi: I see `.optional()` on line 45, so should be okay I think
just wondering if you mean `.allow(null)` vs `.allow('')`

pritamstyz4ever (approved):

config/settings.js
banusha2: We want user to be able change back aliasname to empty string, so I believe `.allow('')` will be good, what do you think on this?

tkyi (approved):

config/settings.js
tkyi: sure

PR #504, opened 2022-08-22
fix: Display JobNameLength in user settings

banusha2:
## Context

Re-order JobNameLength to user settings.

## Objective

Move JobNameLength from User Preference to user settings .

## License

### DIFF
--- a/config/settings.js
+++ b/config/settings.js
@@ -51,13 +51,13 @@ const SCHEMA_USER_SETTINGS = Joi.object()
     .pattern(
         /\d/,
         Joi.object().keys({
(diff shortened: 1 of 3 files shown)


adong (approved):

adong:
I don't know if we need a migration file for it. If not, what's the impact to the users' who has used `JobNameLength` in their pipeline's level rather than user global level

tkyi (approved):
Shouldn't need a migration file since this isn't a field in models. If it is, it's a blob under settings.

adong:
It seems we do need to 

1) add some migration files for users' who use this feature, 
2) or we can have reset user setting api to make this issue go away.

image

PR #509, opened 2022-12-09
fix: remove require from apiUri

klu909:
## Context

apiUri being required breaks some executor

## Objective

remove required() from apiUri property

## License


jithine:
which ones are breaking ? wasn't it supposed to be required ?

klu909:
> which ones are breaking ? wasn't it supposed to be required ?

ones that don't use queueservice such as executor-docker, but we don't use these repo anymore so maybe i will just change their tests

PR #510, opened 2022-12-21
fix(2767): Add stage setup and teardown [1]

tkyi:
## Context

Stages might want to always run initialization or cleanups needed before jobs in a stage can be executed.

## Objective

This PR adds schema support for stage setup/teardown jobs.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2767

## License


config/base.js
```diff
@@ -37,11 +38,16 @@ const SCHEMA_CHILD_PIPELINES = Joi.object()
...
 const SCHEMA_STAGE = Joi.object()
     .keys({
         description: Joi.string(),
-        jobs: Joi.array().items(Job.jobName).min(0)
+        jobs: Joi.array().items(Job.jobName).min(0),
+        setup: SCHEMA_SETUP_TEARDOWN_JOB,
+        teardown: SCHEMA_SETUP_TEARDOWN_JOB,
+        requires: Job.requires,
+        startFrom: Job.jobName,
+        workflowGraph: WorkflowGraph.workflowGraph
```
sagar1312, line 50: Will stage `workflowGraph` contain edges from `setup` to jobs with in the stage?

config/base.js
```diff
@@ -37,11 +38,16 @@ const SCHEMA_CHILD_PIPELINES = Joi.object()
...
+const SCHEMA_SETUP_TEARDOWN_JOB = Job.job.keys({ requires: Joi.forbidden() });
 const SCHEMA_STAGE = Joi.object()
     .keys({
         description: Joi.string(),
-        jobs: Joi.array().items(Job.jobName).min(0)
+        jobs: Joi.array().items(Job.jobName).min(0),
+        setup: SCHEMA_SETUP_TEARDOWN_JOB,
+        teardown: SCHEMA_SETUP_TEARDOWN_JOB,
+        requires: Job.requires,
+        startFrom: Job.jobName,
```
sagar1312, line 49: Why do we need this field?
I believe stage always starts from `setup` (It is also implicitly created when the setup has not been configured explicitly in screwdriver.yaml).
And jobs that need to be triggered after `setup` is successfully completed should be determined from the workflowGraph.
We could configure to have more than one job to get started in parallel after `setup` is successfully completed.

migrations/20230328-initdb-stageBuilds.js
```diff
@@ -0,0 +1,56 @@
...
+                        primaryKey: true,
+                        type: Sequelize.INTEGER.UNSIGNED
+                    },
+                    stageId: {
+                        type: Sequelize.DOUBLE
+                    },
+                    status: {
+                        type: Sequelize.TEXT
+                    },
+                    groupEventId: {
```
sagar1312, line 26: I believe we will have one snapshot for each event
```suggestion
                    eventId: {
```

config/base.js
```diff
@@ -37,11 +38,16 @@ const SCHEMA_CHILD_PIPELINES = Joi.object()
         startAll: Joi.boolean()
     })
     .unknown(false);
-
+const SCHEMA_SETUP_TEARDOWN_JOB = Job.job.keys({ requires: Joi.forbidden() });
 const SCHEMA_STAGE = Joi.object()
     .keys({
         description: Joi.string(),
-        jobs: Joi.array().items(Job.jobName).min(0)
+        jobs: Joi.array().items(Job.jobName).min(0),
```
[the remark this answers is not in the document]
tkyi: sure, added

config/base.js
jithine: can this handle a scenario like this
Screenshot 2023-05-05 at 11 31 45 AM

config/base.js
tkyi: we can add the setup edges but teardown is trickier

config/job.js
```diff
@@ -168,6 +178,7 @@ const SCHEMA_JOB = Joi.object()
         requires: SCHEMA_REQUIRES,
         secrets: SCHEMA_SECRETS,
         settings: SCHEMA_SETTINGS,
+        stage: SCHEMA_STAGE,
```
sagar1312, line 181: I believe the `jobs` attribute inside `stage` definition in `screwdriver.yaml` is the source of truth to associate a `job` to a `stage`. Having `stage` inside `job` configuration would be redundant and needs additional validation to ensure it point to the correct `stage`.
We should not allow it to be configurable under `job`. 

```suggestion
```

config/regex.js
```diff
@@ -50,7 +50,8 @@ module.exports = {
     INTERNAL_TRIGGER: /^~([\w-]+)$/,
     // Stages can only be named with A-Z,a-z,0-9,-,_
     STAGE_NAME: /^[\w-]+$/,
-
+    // Stage trigger like ~stage@deploy or ~stage@stageName:jobName
+    STAGE_TRIGGER: /^~stage@([\w-]+)(?::([\w-]+))?$/,
```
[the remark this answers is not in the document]
tkyi: I think we weren't planning to have join for stages, so all stages were going to be local
not sure what a remote stage would look like syntax-wise

sagar1312, line 54: `~` in the regex should be optional to support `AND` case
```suggestion
    // Stage trigger like ~stage@deploy or ~stage@stageName:jobName or stage@deploy or stage@stageName:jobName
    STAGE_TRIGGER: /^~?stage@([\w-]+)(?::([\w-]+))?$/,
```

sagar1312 (changes requested):

config/regex.js
tkyi: I think we can revisit join case later

sagar1312: Looks like changes in workflow-parser is contradicting the schema. 

Schema has been configured to allow only `OR` case (i.e. the stage job references in `requires` must begin with `~stage@`).

But, workflow-parser only considers stage job references in `requires` that begins with `stage@` and ignores `~stage@`. Also by convention references with prefix `~` are considered as `OR` case. But for stages references without `~` are considered for `OR` case.
https://github.com/screwdriver-cd/workflow-parser/pull/43/files#r1243954316
https://github.com/screwdriver-cd/workflow-parser/pull/43/files#r1243947120
[Test data](https://github.com/screwdriver-cd/workflow-parser/blob/0f241d8e32e662e6abcbf8221aa7be415f580cab/test/lib/getWorkflow.test.js#L87) also omits `~` prefix
```
           jobs: {
                foo: { requires: ['~commit'] },
                A: { requires: ['foo'] },
                B: { requires: ['foo', 'stage@other:teardown'] },
                C: { requires: ['stage@deploy:setup'], stage: { name: 'deploy', startFrom: true } },
                D: { requires: ['~C'], stage: { name: 'deploy' } },
                E: { requires: ['stage@test:setup'], stage: { name: 'test', startFrom: true } },
                main: { requires: ['stage@other:setup'], stage: { name: 'other', startFrom: true } },
                publish: { requires: ['main'], stage: { name: 'other' } }
            }
```

tkyi: Yeah I think that's why I had the `^(?:~)?(stage@)` regex originally before you asked me to change it. I can have it only allow ~ prefix to keep it consistent. What do you think?

sagar1312: > Yeah I think that's why I had the ^(?:~)?(stage@) regex originally before you asked me to change it.

Could you point me to the comment which suggested to update the regex `^(?:~)?(stage@)` to always expect the prefix `~`.

tkyi: https://github.com/screwdriver-cd/workflow-parser/pull/43#discussion_r1243947120

sagar1312: https://github.com/screwdriver-cd/workflow-parser/pull/43#discussion_r1243947120
That code/comment has no reference to the said original regex `^(?:~)?(stage@)`.

That comment just suggests `~stage@` should be considered as `OR` case (this is the convention already being followed through out for non stage jobs) instead of `stage@` being considered as `OR` case stage jobs.
`stage@` by convention should be considered as `AND` case.

tkyi: sure

config/job.js
```diff
@@ -138,8 +138,14 @@ const SCHEMA_TEMPLATEID = Joi.number()
 const SCHEMA_TRIGGER = sdJoi.string().branchFilter();
 const SCHEMA_INTERNAL_TRIGGER = Joi.string().regex(Regex.INTERNAL_TRIGGER); // ~main, ~jobOne
 const SCHEMA_EXTERNAL_TRIGGER = Joi.string().regex(Regex.EXTERNAL_TRIGGER_ALL).example('~sd@1234:component'); // ~sd@123:main or sd@123:main
+const SCHEMA_STAGE_TRIGGER = Joi.string().regex(Regex.STAGE_TRIGGER); // ~stage@deploy, ~stage@deploy:main
```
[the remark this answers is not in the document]
tkyi: when we add join functionality we can change this

config/regex.js
tkyi: For now we can just have OR support and we will revisit supporting AND for stages after we get this first pass for stages working.

sagar1312 (approved):

PR #513, opened 2023-04-27
Add '.', ':', to IMAGE_ALIAS regex

jaysoffian:
This change will allow for an alias such as `xcode:14.1`. Without this change the alias needs to be something unnatural such as `xcode-14_1`.

The image aliases are essentially only dictionary keys, so it's not clear why their naming need be restricted at all, much less be any more restrictive than the image name itself. However, on the assumption that there is a reason for the naming restriction, this commit take the conservative approach of adding only two additional characters.

## Context

This change will allow for an alias such as `xcode:14.1`. Without this change the alias needs to be something unnatural such as `xcode-14_1`.

## Objective

Less restrict naming on image aliases.

## References

None.

## License

### DIFF
--- a/config/regex.js
+++ b/config/regex.js
@@ -97,8 +97,8 @@ module.exports = {
     // First group: SCM plugin name (e.g. github)
     // Second group: SCM host name (e.g. github.com)
     SCM_CONTEXT: /^([^:]+):([^:]+)$/,
-    // Image aliases can only contain A-Z,a-z,0-9,-,_
-    IMAGE_ALIAS: /^[\w-]+$/,
+    // Image aliases can only contain A-Z,a-z,0-9,-,_,.,:
+    IMAGE_ALIAS: /^[\w-.:]+$/,
     // Valid Events for webhook
     WEBHOOK_EVENT: /^~([\w-]+)$/,
     // Provider region. e.g. us-west-1, ap-northeast-2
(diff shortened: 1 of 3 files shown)


jithine (approved):
LGTM. We might want to double check if having `:` in a yaml key actually works. it's valid yaml though.

jaysoffian:
> LGTM. We might want to double check if having `:` in a yaml key actually works. it's valid yaml though.

YAML mappings use a colon-space (`: `) as the delimiter so it's valid:

```
$ yq -o json <<END
images:
  foo:bar: baz
END
{
  "images": {
    "foo:bar": "baz"
  }
}
```

And you can quote keys if need be:

```
$ yq -o json <<END
images:
  "foo: bar": baz
END
{
  "images": {
    "foo: bar": "baz"
  }
}
```

And of course, YAML is a superset of JSON at the end of the day:

```
$ yq -o json <<END
images: {
  "foo: bar": "baz"
}
END
{
  "images": {
    "foo: bar": "baz"
  }
}
```

I don't know why the original restriction was put there since I believe the image aliases are only ever dictionary keys handled internally by SD, but maybe there's a reason for it beyond my foresight.

tkyi:
I don't know if there was a specific reason why we didn't include `:` and `.`, maybe Jithin can better answer that. For me it just didn't occur to me that people would want that in an alias name
https://github.com/screwdriver-cd/screwdriver/issues/1040

tkyi (approved):

PR #515, opened 2023-05-22
feat: add group to build cluster

pritamstyz4ever:
## Context

Logically group build cluster with default based on the environment. 

## Objective

This PR adds a new column `group` to build cluster schema

## License

### DIFF
--- a/models/buildCluster.js
+++ b/models/buildCluster.js
@@ -31,7 +31,13 @@ const MODEL = {
 
     maintainer: Command.maintainer,
 
-    weightage: Joi.number().min(0).max(100).description('Weight percentage for build cluster').example(20)
+    weightage: Joi.number().min(0).max(100).description('Weight percentage for build cluster').example(20),
+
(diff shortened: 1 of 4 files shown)


models/buildCluster.js
```diff
@@ -31,7 +31,13 @@ const MODEL = {
 
     maintainer: Command.maintainer,
 
-    weightage: Joi.number().min(0).max(100).description('Weight percentage for build cluster').example(20)
+    weightage: Joi.number().min(0).max(100).description('Weight percentage for build cluster').example(20),
+
+    group: Joi.string()
+        .valid('on-prem', 'aws', 'gcp')
```
tkyi, line 37: do we need to make it so restrictive? should it just be any string?

tkyi (approved):
we can leave it as is for now and people can add more group options if they need it in the future

tkyi (approved):
we can leave it as is for now and people can add more group options if they need it in the future

PR #516, opened 2023-05-23
feat: make build cluster group name more relaxed

pritamstyz4ever:
## Context

Build cluster group is restricitve

## Objective

This PR updates the group field to be a string and not limit the group names

## License


tkyi (approved):
Not sure if you can just modify the migration file or if you need to add a new one every time

PR #527, opened 2023-08-21
feat(2911): Adding schema for Job Template Pipeline Report

nathom:
## Context

See https://github.com/screwdriver-cd/screwdriver/issues/2911

## Objective

Adds schema for the pipelineUsage endpoint response.

## License


api/pipelineUsage.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const Joi = require('joi');
+const pipelineId = require('../models/pipeline').base.extract('id');
```
sagar1312, line 4:
```suggestion
const pipelineSchema = require('../models/pipeline').base;
const eventSchema = require('../models/event').base;
```

api/pipelineUsage.js
```diff
@@ -0,0 +1,17 @@
...
+module.exports = {
+    get: Joi.array().items(
+        Joi.object({
+            id: pipelineId,
+            name: Joi.string().required(),
+            scmRepo: Scm.repo,
+            lastRun: Joi.alternatives().try(Joi.string().isoDate(), Joi.allow(null)),
+            admins: Joi.object().pattern(Joi.string(), Joi.boolean()).required()
+        })
+    )
```
sagar1312, line 16:
```suggestion
    output: Joi.array().items(
        Joi.object({
            id: pipelineSchema.extract('id').required(),
            name: pipelineSchema.extract('name').required(),
            scmRepo: pipelineSchema.extract('scmRepo').required(),
            lastRun: Joi.alternatives().try(eventSchema.extract('createTime'), Joi.allow(null)),
            admins: pipelineSchema.extract('admins').required()
        })
    )
```

sagar1312 (changes requested):

.prettierrc.yaml
tkyi, line 1: did you mean to check this in? can you add this file to the `.gitignore` file

api/pipelineUsage.js
tkyi: I think `get` better matches the naming format we've been using for other files.

.prettierrc.yaml
nathom: I think it is useful because the style rules that screwdriver adheres to are different from the prettier defaults. This means that anyone who has format-on-write set up is going to fail lint by default. But if its not appropriate to keep, I can remove it.

sagar1312: Instead of copying the file to each module repo, can we centralize and share the configuration?
https://prettier.io/docs/en/configuration.html#sharing-configurations

test/api/pipelineUsage.test.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+const { assert } = require('chai');
+const { api } = require('../..');
+const { validate } = require('../helper');
+
+describe('api get pipelineUsage', () => {
+    const inputSchema = api.pipelineUsage.get;
```
sagar1312, line 8: This is not input schema
```suggestion
    const schema = api.pipelineUsage.get;
```

test/api/pipelineUsage.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+    describe('basic', () => {
```
sagar1312, line 10: This segregation may not be need if all the tests are under one `describe`
```suggestion
    
```

sagar1312 (changes requested):
This PR needs rebase to catch up with the `master` branch

sagar1312 (approved):

PR #529, opened 2023-09-07
fix(2767): Job permutation for stage in validator

tkyi:
## Context

Need to fix syntax for job permutation, should only take one stage.

## Objective

This PR changes job permutation validation field from `stages` to `stage`.

## References

Related to https://github.com/screwdriver-cd/config-parser/pull/143#discussion_r1312135136

## License


api/validator.js
```diff
@@ -36,7 +36,7 @@ const SCHEMA_JOB_PERMUTATION = Joi.object()
         secrets: Job.secrets,
         settings: Job.settings,
         sourcePaths: Job.sourcePaths,
-        stages: Base.stages,
+        stage: Base.stage,
```
sagar1312, line 39: Do we need to allow the entire stage object to be stored in a job permutation?
```
const SCHEMA_STAGE = Joi.object()
    .keys({
        description: Joi.string(),
        jobs: Joi.array().items(Job.jobName).min(0),
        setup: SCHEMA_SETUP_TEARDOWN_JOB,
        teardown: SCHEMA_SETUP_TEARDOWN_JOB,
        requires: Job.requires
    })
    .unknown(false);
```

I believe we only need stage `name` (or `id`) for reverse lookup per [this change](https://github.com/screwdriver-cd/config-parser/pull/143/files#diff-650a22c636f250a0b20166054e7ffaa7a337afe42b97be42acc2728c96107d0aR564).
```suggestion
        stage: Joi.object({
            name:  Joi.string().regex(Regex.STAGE_NAME).required()
        }).keys().unknown(false),
```

sagar1312 (changes requested):
Please add tests to cover this.

api/validator.js
```diff
@@ -36,7 +36,11 @@ const SCHEMA_JOB_PERMUTATION = Joi.object()
         secrets: Job.secrets,
         settings: Job.settings,
         sourcePaths: Job.sourcePaths,
-        stages: Base.stages,
+        stage: Joi.object()
+            .keys({
+                name: Joi.string().regex(Regex.STAGE_NAME).optional()
+            })
+            .unknown(false),
```
sagar1312, line 43: Having `stage` should be optional. If stage object exists, it must have `name`
```suggestion
        stage: Joi.object()
            .keys({
                name: Joi.string().regex(Regex.STAGE_NAME).required()
            })
            .optional()
            .unknown(false),
```

tkyi: was troubleshooting something for config-parser and forgot to change it back

sagar1312 (approved):

PR #531, opened 2023-10-19
feat: add sonar badge to pipeline

adong:
## Context

To add badges, and include one example of `sonar` to the pipeline, so we can update the sonar dashboard name and uri in the pipeline option page.

## Objective
image

image

image

## License

### DIFF
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -37,7 +37,11 @@ const CREATE_MODEL = {
 };
 
 const MODEL = {
-    id: Joi.number().integer().positive().description('Identifier of this pipeline').example(123345),
+    id: Joi.number()
+        .integer()
+        .positive()
+        .description('Identifier of this pipeline')
+        .example(123345),
 
     name: Scm.repoName.optional(),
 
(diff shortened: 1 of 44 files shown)


models/pipeline.js
```diff
@@ -90,7 +90,19 @@ const MODEL = {
                 actions: Joi.array().items(Joi.string())
             })
         )
-        .description('List of subscribed scm urls paired with actions')
+        .description('List of subscribed scm urls paired with actions'),
+
+    sonarDashboardName: Joi.string()
```
tkyi, line 95: do you need to add these to `get` (line 133)?

jithine, line 95: what if we have more integrations in the future ? Do we need a flat structure like this ? ie keep adding more fields for potential integrations like Saucelabs Project UI, Mend UI, Semgrep etc ?

adong: Updated

models/pipeline.js
```diff
@@ -11,6 +11,16 @@ const Parameters = require('../config/parameters');
 const mutate = require('../lib/mutate');
 
 const STATES = ['ACTIVE', 'INACTIVE'];
+const BADGE = Joi.object({
```
tkyi, line 14:
```suggestion
const SONAR = Joi.object({
```

adong: Oh, this is not a typo. This is written for extendibility, so we might be able to add something like in the future.

```
{
    sonar: BADGE,
    saucelabs: BADGE
}
```

models/pipeline.js
```diff
@@ -90,7 +97,14 @@ const MODEL = {
...
             })
         )
-        .description('List of subscribed scm urls paired with actions')
+        .description('List of subscribed scm urls paired with actions'),
+
+    badges: Joi.object({
+        sonar: BADGE
+    })
+        .description('A list of badges that pipline has')
+        .example(`sonar: { name: 'dashboard', uri: 'http://sonar.com/sample1'}`)
```
tkyi, line 106:
```suggestion
        .example(`{ sonar: { name: 'dashboard', uri: 'http://sonar.com/sample1' } }`)
```

tkyi (approved):

PR #539, opened 2023-10-27
fix(2767): Relax job name Regex to allow stage setup/teardown jobs

sagar1312:
## Context
Regex associated with `job` name is strict and does not allow setup and teardown jobs created by the backend for a stage which follows the pattern like `stage@<STAGE_NAME>:setup` or `stage@<STAGE_NAME>:teardown`
Ex: `stage@production:setup`

This results in below issues:

1. Validator fails to permit the pipeline configuration which has one or more stages defined
image

2. Validation of the API response for the endpoint `/v4/pipelines/<PIPELINE_ID>/jobs` fails when stage setup/teardown jobs exist. 
image

## Objective

Relaxing the job name Regex to allow stage setup/teardown jobs

## References

https://github.com/screwdriver-cd/screwdriver/issues/2767

## License

### DIFF
--- a/config/regex.js
+++ b/config/regex.js
@@ -41,7 +41,8 @@ module.exports = {
     // Steps can only be named with A-Z,a-z,0-9,-,_
     STEP_NAME: /^[\w-]+$/,
     // Jobs can only be named with A-Z,a-z,0-9,-,_
-    JOB_NAME: /^[\w-]+$/,
+    // Also allow stage setup/teardown like stage@stage-name:setup
+    JOB_NAME: /^(([\w-]+)|(stage@[\w-]+:(setup|teardown)))$/,
     // PR JOB Name can only be PR-1 or PR-1:main, group1: PR-prNum, group2: jobName
     PR_JOB_NAME: /^(PR-\d+)(?::([\w-]+))?$/,
     // Match all possible job name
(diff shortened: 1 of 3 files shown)


config/regex.js
```diff
@@ -41,7 +41,8 @@ module.exports = {
     // Steps can only be named with A-Z,a-z,0-9,-,_
     STEP_NAME: /^[\w-]+$/,
     // Jobs can only be named with A-Z,a-z,0-9,-,_
-    JOB_NAME: /^[\w-]+$/,
+    // Also allow stage setup/teardown like stage@stage-name:setup
+    JOB_NAME: /^(([\w-]+)|(stage@[\w-]+:(setup|teardown)))$/,
```
tkyi, line 45: wondering if something like this would help with matching/grouping 
```suggestion
    JOB_NAME: /^(([\w-]+)|(?:stage@([\w-]+):(setup|teardown)))$/,
```
Screenshot 2023-10-27 at 3 23 57 PM

models/job.js
```diff
@@ -9,7 +9,7 @@ const MODEL = {
     id: Joi.number().integer().positive().description('Identifier of this Job').example(123345),
 
     name: Joi.string()
-        .regex(/^(PR-[0-9]+:)?[\w-]+$/)
+        .regex(/^(PR-[0-9]+:)?(([\w-]+)|(stage@[\w-]+:(setup|teardown)))$/)
```
tkyi, line 12: similar suggestion to above

sagar1312: Done

tkyi (approved):

PR #547, opened 2024-01-22
fix(2767): Stage must always have valid setup and teardown job IDs

sagar1312:
## Context

`setup` and `teardown` should always exist for a stage.
A stage must have at least one job.

## Objective

Update model and DB schema
- `setup` and `teardown` are non nullable
- `jobIds` should not be empty

## References

https://github.com/screwdriver-cd/screwdriver/issues/276

## License

### DIFF
--- a/models/stage.js
+++ b/models/stage.js
@@ -11,12 +11,13 @@ const MODEL = {
 
     name: Joi.string().regex(Regex.STAGE_NAME).max(110).description('Name of the Stage').example('deploy'),
 
-    setup: Joi.number().integer().positive().description('Job ID for Stage setup').example(123345),
+    setup: Joi.number().integer().positive().required().description('Job ID for Stage setup').example(123345),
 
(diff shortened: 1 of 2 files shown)


jithine (approved):

models/stage.js
```diff
@@ -11,12 +11,13 @@ const MODEL = {
 
     name: Joi.string().regex(Regex.STAGE_NAME).max(110).description('Name of the Stage').example('deploy'),
 
-    setup: Joi.number().integer().positive().description('Job ID for Stage setup').example(123345),
+    setup: Joi.number().integer().positive().required().description('Job ID for Stage setup').example(123345),
 
-    teardown: Joi.number().integer().positive().description('Job ID for Stage teardown').example(123345),
+    teardown: Joi.number().integer().positive().required().description('Job ID for Stage teardown').example(123345),
```
adong, line 16: How about adding `default()`?

PR #551, opened 2024-02-22
feat(3020): change to support pipeline template

klu909:
## Context

Add `template` field and mark currently unsupported field as forbidden when using pipeline template
## References

https://github.com/screwdriver-cd/screwdriver/issues/3020
## License

### DIFF
--- a/config/base.js
+++ b/config/base.js
@@ -65,15 +65,24 @@ const SCHEMA_SUBSCRIBE = Joi.object().keys({
 
 const SCHEMA_CONFIG = Joi.object()
     .keys({
+        template: Joi.string().regex(Regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE),
         version: Joi.number().integer().min(1).max(50),
-        annotations: Annotations.annotations,
-        jobs: SCHEMA_JOBS.required(),
-        shared: SCHEMA_SHARED,
-        cache: SCHEMA_CACHE,
-        childPipelines: SCHEMA_CHILD_PIPELINES,
-        stages: SCHEMA_STAGES,
-        subscribe: SCHEMA_SUBSCRIBE,
-        parameters: Parameters.parameters.default({})
+        annotations: Annotations.annotations.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
+        jobs: SCHEMA_JOBS.when('template', { is: Joi.exist(), then: Joi.forbidden(), otherwise: Joi.required() }),
+        shared: Joi.when('template', {
+            is: Joi.exist(),
+            then: Joi.object().keys({
+                image: Job.image,
+                environment: Job.environment,
+                settings: Job.settings
+            }),
+            otherwise: SCHEMA_SHARED
+        }),
+        cache: SCHEMA_CACHE.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
+        childPipelines: SCHEMA_CHILD_PIPELINES.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
+        stages: SCHEMA_STAGES.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
+        subscribe: SCHEMA_SUBSCRIBE.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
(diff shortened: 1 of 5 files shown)


test/config/base.test.js
```diff
@@ -44,4 +44,10 @@ describe('config base', () => {
             assert.isNull(validate('config.base.stages.yaml', config.base.stages).error);
         });
     });
+
+    describe.only('pipeline template', () => {
```
tkyi, line 48:
```suggestion
    describe('pipeline template', () => {
```

config/base.js
```diff
@@ -65,9 +65,10 @@ const SCHEMA_SUBSCRIBE = Joi.object().keys({
 
 const SCHEMA_CONFIG = Joi.object()
     .keys({
+        template: Joi.string().regex(Regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE),
         version: Joi.number().integer().min(1).max(50),
         annotations: Annotations.annotations,
-        jobs: SCHEMA_JOBS.required(),
+        jobs: SCHEMA_JOBS.optional(),
```
tkyi, line 71: should we be making this optional or make a separate schema config for pipeline template?

klu909: pipeline that end up with no jobs will still be caught by config-parser, unless there's something else that uses this schema I would say we don't have to?

sagar1312: We should conditionally make `jobs` as `required` in the schema when `template` is not specified

config/base.js
```diff
@@ -65,9 +65,10 @@ const SCHEMA_SUBSCRIBE = Joi.object().keys({
 
 const SCHEMA_CONFIG = Joi.object()
     .keys({
+        template: Joi.string().regex(Regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE),
         version: Joi.number().integer().min(1).max(50),
         annotations: Annotations.annotations,
-        jobs: SCHEMA_JOBS.required(),
+        jobs: SCHEMA_JOBS.when('template', { not: Joi.exist(), then: Joi.required(), otherwise: Joi.optional() }),
```
sagar1312, line 71: Shouldn't we disallow `jobs` for initial version of `Pipeline Templates` feature?
Also extend the similar schema changes to other fields as per the design?
```suggestion
        jobs: SCHEMA_JOBS.when('template', { not: Joi.exist(), then: Joi.required(), otherwise: Joi.forbidden() }),
```

sagar1312 (changes requested):

config/base.js
klu909: it will just be discarded in config-parser

config/base.js
```diff
@@ -65,15 +65,24 @@ const SCHEMA_SUBSCRIBE = Joi.object().keys({
...
-        subscribe: SCHEMA_SUBSCRIBE,
-        parameters: Parameters.parameters.default({})
+        annotations: Annotations.annotations.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
+        jobs: SCHEMA_JOBS.when('template', { is: Joi.exist(), then: Joi.forbidden(), otherwise: Joi.required() }),
+        shared: Joi.when('template', {
+            is: Joi.exist(),
+            then: Joi.object().keys({
+                image: Job.image,
+                environment: Job.environment,
+                settings: Job.settings
```
sagar1312, line 77: This could be more intuitive that we are allowing subset of `SCHEMA_SHARED` schema.
```suggestion
                image: SCHEMA_SHARED.image,
                environment: SCHEMA_SHARED.environment,
                settings: SCHEMA_SHARED.settings
```

sagar1312 (changes requested):

config/base.js
klu909: unfortunately it doesn't work, that's first thing i tried, it would throw error

sagar1312: you could try 
```
   image: SCHEMA_SHARED.extract('image'),
   environment: SCHEMA_SHARED.extract('environment'),
   settings: SCHEMA_SHARED.extract('settings')
```

sagar1312 (approved):

PR #552, opened 2024-02-26
feat(3022): Add annotation to mark implicitly created stage setup/teardown jobs are virtual

sagar1312:
## Context
`setup` / `teardown` job for a `stage` are implicitly created if it is not configured in the `screwdriver.yaml`.
We need a mechanism to identify these implicitly created `setup` / `teardown` jobs to take necessary actions.
Ex: skip execution, hide it from workflow graph in the UI, etc

## Objective

This PR adds a new annotation to job annotations: `screwdriver.cd/virtual`.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3022

## License

### DIFF
--- a/config/annotations.js
+++ b/config/annotations.js
@@ -26,7 +26,8 @@ const RESERVED_JOB_ANNOTATIONS = [
     'screwdriver.cd/manualStartEnabled',
     'screwdriver.cd/blockedBySameJob',
     'screwdriver.cd/blockedBySameJobWaitTime',
-    'screwdriver.cd/jobDisabledByDefault'
+    'screwdriver.cd/jobDisabledByDefault',
+    'screwdriver.cd/virtual'
 ];
 const RESERVED_PIPELINE_ANNOTATIONS = [
     'screwdriver.cd/buildCluster',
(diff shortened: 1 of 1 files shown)


adong (dismissed):

config/annotations.js
```diff
@@ -26,7 +26,8 @@ const RESERVED_JOB_ANNOTATIONS = [
     'screwdriver.cd/manualStartEnabled',
     'screwdriver.cd/blockedBySameJob',
     'screwdriver.cd/blockedBySameJobWaitTime',
-    'screwdriver.cd/jobDisabledByDefault'
+    'screwdriver.cd/jobDisabledByDefault',
+    'screwdriver.cd/virtual'
```
adong, line 30: Would the name `virtualTrigger` make more sense?

tkyi: I think virtualTrigger would be more clear

sagar1312: Done. Changed to `virtualTrigger`.

jithine: Intention is that the job is a virtual job. Where does trigger come into picture ?

PR #554, opened 2024-02-28
feat: add column templateVersionId to pipeline

klu909:
## Context

add column templateVersionId to pipeline
## Objective

add column templateVersionId to pipeline
## License

### DIFF
--- a/config/base.js
+++ b/config/base.js
@@ -63,7 +63,7 @@ const SCHEMA_SUBSCRIBE = Joi.object().keys({
     )
 });
 
-const SCHEMA_CONFIG = Joi.object()
+const SCHEMA_MERGE_CONFIG = Joi.object()
     .keys({
         template: Joi.string().regex(Regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE),
         version: Joi.number().integer().min(1).max(50),
@@ -82,7 +82,22 @@ const SCHEMA_CONFIG = Joi.object()
         childPipelines: SCHEMA_CHILD_PIPELINES.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
         stages: SCHEMA_STAGES.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
         subscribe: SCHEMA_SUBSCRIBE.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
-        parameters: Parameters.parameters.default({}).when('template', { is: Joi.exist(), then: Joi.forbidden() }),
+        parameters: Parameters.parameters.default({}).when('template', { is: Joi.exist(), then: Joi.forbidden() })
+    })
+    .unknown(false);
+
+const SCHEMA_CONFIG = Joi.object()
+    .keys({
+        template: Joi.string().regex(Regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE),
+        version: Joi.number().integer().min(1).max(50),
+        annotations: Annotations.annotations,
+        jobs: SCHEMA_JOBS.required(),
+        shared: SCHEMA_SHARED,
+        cache: SCHEMA_CACHE,
+        childPipelines: SCHEMA_CHILD_PIPELINES,
+        stages: SCHEMA_STAGES,
+        subscribe: SCHEMA_SUBSCRIBE,
+        parameters: Parameters.parameters.default({}),
         templateVersionId: Joi.number().integer().positive().optional().allow(null)
     })
     .unknown(false);
@@ -98,6 +113,7 @@ module.exports = {
     cache: SCHEMA_CACHE,
     cachePerm: SCHEMA_CACHE_PERMUTATION,
     childPipelines: SCHEMA_CHILD_PIPELINES,
+    mergeConfig: SCHEMA_MERGE_CONFIG,
     config: SCHEMA_CONFIG,
     stageSetupTeardownJob: SCHEMA_SETUP_TEARDOWN_JOB,
     stage: SCHEMA_STAGE,
--- a/config/base.js
+++ b/config/base.js
@@ -63,7 +63,7 @@ const SCHEMA_SUBSCRIBE = Joi.object().keys({
     )
 });
 
-const SCHEMA_MERGE_CONFIG = Joi.object()
+const SCHEMA_CONFIG_PRE_TEMPLATE_MERGE = Joi.object()
     .keys({
         template: Joi.string().regex(Regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE),
         version: Joi.number().integer().min(1).max(50),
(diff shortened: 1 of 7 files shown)


models/pipeline.js
```diff
@@ -112,7 +112,13 @@ const MODEL = {
...
         .description('A list of badges that pipline has')
         .example(`{ sonar: { name: 'dashboard', uri: 'http://sonar.com/sample1' } }`)
-        .optional()
+        .optional(),
+
+    templateVersionId: Joi.number()
+        .integer()
+        .positive()
+        .description("Identifier for this pipeline's template")
+        .example(123345)
```
VonnyJap, line 121: do we need `optional` here? this is not mandatory field for pipeline object model, I believe.

klu909: it was used for get validation thus we don't really have to set optional explicitly because it will always be there, but yeah we can add it anyway

config/base.js
```diff
@@ -82,7 +82,8 @@ const SCHEMA_CONFIG = Joi.object()
         childPipelines: SCHEMA_CHILD_PIPELINES.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
         stages: SCHEMA_STAGES.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
         subscribe: SCHEMA_SUBSCRIBE.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
-        parameters: Parameters.parameters.default({}).when('template', { is: Joi.exist(), then: Joi.forbidden() })
+        parameters: Parameters.parameters.default({}).when('template', { is: Joi.exist(), then: Joi.forbidden() }),
+        templateVersionId: Joi.number().integer().positive().optional().allow(null)
```
sagar1312, line 86: Is `templateVersionId` a valid field to have on the config?
This this shouldn't be allowed to be specified in `screwdriver.yaml`.

migrations/20240228-add_column_templateVersionId_to_pipeline.JS
```diff
@@ -0,0 +1,26 @@
...
+const table = `${prefix}pipelines`;
+
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        await queryInterface.sequelize.transaction(async transaction => {
+            await queryInterface.addColumn(
+                table,
+                'templateVersionId',
+                {
+                    type: Sequelize.INTEGER
```
sagar1312, line 15: To keep it consistent
https://github.com/screwdriver-cd/data-schema/blob/master/migrations/20230727160440-initdb-pipelineTemplateVersions.js#L19
```suggestion
                    type: Sequelize.INTEGER.UNSIGNED
```

migrations/20240228-add_column_templateVersionId_to_pipeline.JS
```diff
@@ -0,0 +1,26 @@
...
+                },
+                { transaction }
+            );
+
+            await queryInterface.addIndex(table, ['templateVersionId'], {
+                name: `${table}_template_version_id`,
+                transaction
+            });
```
sagar1312, line 23: Per doc, it is suggested to use `options.fields`
https://sequelize.org/api/v6/class/src/dialects/abstract/query-interface.js~queryinterface#instance-method-addIndex
```suggestion
            await queryInterface.addIndex(table, {
                name: `${table}_template_version_id`,
                fields: ['templateVersionId'],
                transaction
            });
```

models/pipeline.js
sagar1312: Schema would be optional by default

sagar1312 (changes requested):

config/base.js
klu909: it's needed because it will be added in config parser after we merge the template

klu909: since it's a field that purely just provides information to user, i think there's any harm. do you have any suggestion?

sagar1312: This could potentially be an input from the user even though it is not intended/recommended by design. If we add this to config schema, it would mean that `templateVersionId` is a valid field to be specified in the pipeline configuration `screwdriver.yaml`.

Suggestions:
1. Have 2 variations of the schema: one with `template` and other with `templateId`. Use the first one validation before merging pipeline template and second one for validation post merge.
2. Make `template` and `templateId` at least mutually exclusive. Leave a note for `templateId` that it is not expected to be configured in `screwdriver.yaml`.

I would incline towards option 1 as it is more cleaner and completely blocks the use from specifying it in `screwdriver.yaml`.

Considering the same problem exists for job template, option 2 could be considered for short term.

config/base.js
```diff
@@ -86,6 +86,22 @@ const SCHEMA_CONFIG = Joi.object()
     })
     .unknown(false);
 
+const SCHEMA_CONFIG = Joi.object()
+    .keys({
+        template: Joi.string().regex(Regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE),
```
sagar1312, line 91: `template` would be removed while merging pipeline template
```suggestion
```

config/base.js
```diff
@@ -86,6 +86,22 @@ const SCHEMA_CONFIG = Joi.object()
     })
     .unknown(false);
 
+const SCHEMA_CONFIG = Joi.object()
```
sagar1312, line 89: Consider renaming. Keeping it generic would be ambiguous.
```suggestion
const SCHEMA_CONFIG_POST_TEMPLATE_MERGE = Joi.object()
```

config/base.js
```diff
@@ -63,7 +63,7 @@ const SCHEMA_SUBSCRIBE = Joi.object().keys({
     )
 });
 
-const SCHEMA_CONFIG = Joi.object()
+const SCHEMA_MERGE_CONFIG = Joi.object()
```
sagar1312, line 66: Consider renaming. Just having `MERGE` could be confusing. It is not clear whether it is applied before merge, after merge and during merge, etc
```suggestion
const SCHEMA_CONFIG_PRE_TEMPLATE_MERGE = Joi.object()
```

config/base.js
```diff
@@ -97,6 +113,7 @@ module.exports = {
     cache: SCHEMA_CACHE,
     cachePerm: SCHEMA_CACHE_PERMUTATION,
     childPipelines: SCHEMA_CHILD_PIPELINES,
+    mergeConfig: SCHEMA_MERGE_CONFIG,
     config: SCHEMA_CONFIG,
```
sagar1312, line 117: Having one generic and one specific names could be confusing. Let us use more precise names for both
```suggestion
    configBeforeMergingTemplate: SCHEMA_CONFIG_PRE_TEMPLATE_MERGE,
    configAfterMergingTemplate: SCHEMA_CONFIG_POST_TEMPLATE_MERGE,
```

sagar1312 (changes requested):

config/base.js
```diff
@@ -97,7 +112,8 @@ module.exports = {
     cache: SCHEMA_CACHE,
     cachePerm: SCHEMA_CACHE_PERMUTATION,
     childPipelines: SCHEMA_CHILD_PIPELINES,
-    config: SCHEMA_CONFIG,
+    configBeforeMergingTemplate: SCHEMA_CONFIG_PRE_TEMPLATE_MERGE,
```
sagar1312, line 115: This is a breaking change. Please follow https://www.npmjs.com/package/semantic-release to bump up `major` version

sagar1312 (approved):
Looks good

PR #560, opened 2024-03-14
feat(3059): Add pipeline level customization for pipeline template

klu909:
## Context

relax restrictions for pipeline level setting in screwdriver yaml that uses pipeline template
## Objective

relax restrictions for pipeline level setting in screwdriver yaml that uses pipeline template
## References
https://github.com/screwdriver-cd/screwdriver/issues/3058

## License

### DIFF
--- a/config/base.js
+++ b/config/base.js
@@ -71,9 +71,10 @@ const SCHEMA_CONFIG_PRE_TEMPLATE_MERGE = Joi.object()
         jobs: SCHEMA_JOBS.when('template', { is: Joi.exist(), then: Joi.forbidden(), otherwise: Joi.required() }),
         shared: Joi.when('template', {
             is: Joi.exist(),
-            then: SCHEMA_SHARED.append({
-                order: Joi.forbidden(),
-                steps: Joi.forbidden()
+            then: Joi.object().keys({
+                image: Job.image,
+                environment: Job.environment,
+                settings: Job.settings
             }),
             otherwise: SCHEMA_SHARED
         }),
--- a/config/base.js
+++ b/config/base.js
@@ -67,22 +67,21 @@ const SCHEMA_CONFIG_PRE_TEMPLATE_MERGE = Joi.object()
     .keys({
         template: Joi.string().regex(Regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE),
         version: Joi.number().integer().min(1).max(50),
-        annotations: Annotations.annotations.when('template', { is: Joi.exist(), then: Joi.forbidden() }),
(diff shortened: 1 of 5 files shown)


config/base.js
```diff
@@ -67,22 +67,21 @@ const SCHEMA_CONFIG_PRE_TEMPLATE_MERGE = Joi.object()
...
         jobs: SCHEMA_JOBS.when('template', { is: Joi.exist(), then: Joi.forbidden(), otherwise: Joi.required() }),
         shared: Joi.when('template', {
             is: Joi.exist(),
-            then: Joi.object().keys({
-                image: Job.image,
-                environment: Job.environment,
-                settings: Job.settings
+            then: SCHEMA_SHARED.append({
+                order: Joi.forbidden(),
+                steps: Joi.forbidden()
```
sagar1312, line 76: Since we are not yet allowing customization of steps of the jobs inherited from the template and also not supporting defining new jobs, we should continue to selectively allow the fields that are supported. We should not allow customization of any field that changes the behavior of the jobs inherited from the template for now.

```suggestion
            then: Joi.object().keys({
                image: Job.image,
                environment: Job.environment,
                settings: Job.settings
```

sagar1312 (changes requested):
PR title and description is very generic. We can expect more PRs in future around Pipeline Template enhancements.

Can we have the title and description updated to convey what is the enhancement being made in this PR?

config/base.js
klu909: that was literally being asked from the issue `Support customization for other shared configs: annotations, blockedBy, cache, description, freezeWindows, parameters, provider, requires, secrets, sourcePaths, template`

sagar1312: The task is to support customization for the specified fields in a broader sense. It does not imply that we should relax the schema to allow all these at once.

We need to analyze each field individually and define how the customization should be interpreted for various scenarios. Implement the logic to handle it in the backend before relaxing the schema.

test/data/config.pipelineTemplate.yaml
```diff
@@ -6,7 +6,69 @@ maintainer: foo@bar.com
...
-                -   test: npm test
+                - hello: echo hello
+    shared:
+        image: node:20
+        environment:
+            FOO: user overwrite
+        settings:
+            email: foo@example.com
+        blockedBy: [~main]
+        cache: true
```
sagar1312, line 22: This may not be a valid field in config 
```suggestion
```

sagar1312 (changes requested):

sagar1312 (approved):

PR #561, opened 2024-04-03
feat: add example to use parameters for event model

VonnyJap:
## Context

A client can add parameters as payload in the meta field when creating an event via `/v4/events` POST API. But the API documentation use `{}` empty object as the default `meta` payload. 

## Objective

Provide a concise example for clients to send parameters payload when creating an event through API. 

## License

### DIFF
--- a/models/event.js
+++ b/models/event.js
@@ -35,7 +35,7 @@ const MODEL = {
     meta: Joi.object()
         .default({})
         .description('Key=>Value information from the event itself')
-        .example({parameters: { nameA: 'value1', jobA: { nameB: 'value2'}}}),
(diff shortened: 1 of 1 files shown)


models/event.js
```diff
@@ -32,7 +32,12 @@ const MODEL = {
         .description('When this event was created')
         .example('2038-01-19T03:14:08.131Z'),
     creator: Scm.user.description('Creator of the event'),
-    meta: Joi.object().default({}).description('Key=>Value information from the event itself'),
+    meta: Joi.object().description('Key=>Value information from the event itself').example({
```
tkyi, line 35: probably would be good to leave the default value and just add an example?

jithine (approved):

PR #585, opened 2024-10-14
feat(3211): Add a new field to builds to indicate the severity of the status message

sagar1312:
## Context

All the build status message are currently treated/visualized as warning in the UI. We need a mechanism to distinguish the messages based on severity

## Objective

Add a new field `statusMessageType` to the `build` with `ERROR`, `WARN` and `INFO` as the allowed values

## References

https://github.com/screwdriver-cd/screwdriver/issues/3211

## License

### DIFF
--- a/models/build.js
+++ b/models/build.js
@@ -22,6 +22,7 @@ const STATUSES = [
     'COLLAPSED', // when the build is collapsed
     'FROZEN' // when the build is frozen due to freeze window
 ];
+const STATUS_MESSAGE_TYPES = ['ERROR', 'WARN', 'INFO'];
 
 const MODEL = {
     id: Joi.number().integer().positive().description('Identifier of this build').example(123345),
@@ -84,6 +85,12 @@ const MODEL = {
     statusMessage: Joi.string()
         .description('Status message to describe status of the build')
         .example('Build failed due to infrastructure error'),
+
+    statusMessageType: Joi.string()
+        .valid(...STATUS_MESSAGE_TYPES)
+        .max(10)
+        .description('Severity of the status message')
+        .example('WARN'),
 
     stats: Joi.object()
         .keys({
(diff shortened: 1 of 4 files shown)


models/build.js
```diff
@@ -22,6 +22,7 @@ const STATUSES = [
     'COLLAPSED', // when the build is collapsed
     'FROZEN' // when the build is frozen due to freeze window
 ];
+const STATUS_MESSAGE_TYPES = ['ERROR', 'WARN', 'INFO'];
```
tkyi, line 25:
```suggestion
const STATUS_MESSAGE_TYPES = ['ERR', 'WARN', 'INFO'];
```
since the other two are shortened, wondering if this is more consistent

models/build.js
```diff
@@ -85,6 +86,12 @@ const MODEL = {
         .description('Status message to describe status of the build')
         .example('Build failed due to infrastructure error'),
 
+    statusMessageType: Joi.string()
+        .valid(...STATUS_MESSAGE_TYPES)
```
tkyi, line 90: should we set `default`?

models/build.js
sagar1312: We don't need to shorten `ERROR`. The values used are pretty much the industry standard.
Ex:
https://github.com/winstonjs/winston?tab=readme-ov-file#using-logging-levels
https://www.tutorialspoint.com/log4j/log4j_logging_levels.htm

models/build.js
sagar1312: This field is optional/nullable. Not all builds have `statusMessage` set. So, is the `statusMessageType`.
We don't need to assign a default value.

tkyi (commented):
Can you add new field to example?
https://github.com/screwdriver-cd/data-schema/blob/master/test/data/build.update.yaml

tkyi (approved):

PR #594, opened 2025-03-18
fix(3310): Consistently extract external job name

sagar1312:
## Context
The regex `^~?sd@(\d+):(?:([\w-]+)|(stage@([\w-]+):setup))$` does not consistently return the matching group to get the job name 

image

## Objective

Fix the regex to consistently retrieve the job name as the second group.

image

## References

https://github.com/screwdriver-cd/screwdriver/issues/3310

## License

### DIFF
--- a/test/config/regex.test.js
+++ b/test/config/regex.test.js
@@ -549,4 +549,53 @@ describe('config regex', () => {
             });
         });
     });
+
+    describe.only('extract downstream job name', () => {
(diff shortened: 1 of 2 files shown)


test/config/regex.test.js
```diff
@@ -549,4 +549,53 @@ describe('config regex', () => {
             });
         });
     });
+
+    describe.only('extract downstream job name', () => {
```
jithine, line 553: pls remove only

jithine (approved):

PR #595, opened 2025-03-24
feat(3312): Allow pipeline to have users with admin privileges outside the pipeline the SCM context

sagar1312:
## Context

`admins` column in the `pipelines` table holds SCM usernames with admin privileges within the pipeline SCM context.

Sample pipeline: 
| id | scmContext | admins |
| :--- | :--- | :--- |
| 41 | github:github.com | {"sagar1312":true, "sam": false} |

Sample users:
| id | username | scmContext |
| :--- | :--- | :--- |
| 1 | sagar1312 | github:github.com |
| 2 | sam | github:github.com |
| 3 | sagar1312 | bitbucket:bitbucket.com |
| 4 | bob | bitbucket:bitbucket.com |

Using SCM username as the identifier will limit us from resolving the conflict when there are multiple users with same SCM username but associated with different SCMs.
For Ex: There are 2 users with same SCM username "sagar1312", one in GitHub and the other in Bitbucket.

## Objective
We need a mechanism to store users with admin privileges for a pipeline who is not part of the pipeline SCM context.

Add a new column `adminUserIds` to `pipeline` table to store Screwdriver user identifiers (`id` from `users` table). This allows us to store users outside the pipeline SCM context.

Sample pipeline: 
| id | scmContext | admins | adminUserIds |
| :--- | :--- | :--- | :--- |
| 41 | github:github.com | {"sagar1312":true, "sam": false} | [1, 3] |

P.S. 
We will have to retain `admins` column for backward compatibility.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3312

## License


migrations/20211215-add_column_parameters_to_templates.js
```diff
@@ -12,7 +12,9 @@ module.exports = {
                 table,
                 'parameters',
                 {
-                    type: Sequelize.TEXT('medium')
+                    type: Sequelize.TEXT('medium'),
+                    defaultValue: '[]',
+                    allowNull: false
```
VonnyJap, line 17: is this change intentional? this is an old migration file and it belongs to a template table which is aside from the change required on this PR?

VonnyJap:
I did not see a migration file to add new column, was it an overlooked?

migrations/20211215-add_column_parameters_to_templates.js
sagar1312: Reverted.

sagar1312:
> I did not see a migration file to add new column, was it an overlooked?

Git ignored the new file during commit. Added it now.

VonnyJap (approved):

PR #596, opened 2025-04-18
fix: Relax pipeline adminUserIds restriction

kumada626:
## Context

If screwdriver using mysql 5.X, default value cannot be set in TEXT field.
Because of that, old pipelines have null adminUserIds column.

## Objective

Relax pipeline adminUserIds restriction for mysql backward compatibility.
- remove required restriction for null value.
- move adminUserIds from requiredFileds to optionalFields.

## References

https://github.com/screwdriver-cd/screwdriver/pull/3325

## License

### DIFF
--- a/models/pipeline.js
+++ b/models/pipeline.js
@@ -71,8 +71,8 @@ const MODEL = {
     adminUserIds: Joi.array()
         .items(Joi.number().integer().positive().description('Identifier of the user').example(12345))
         .description('IDs of the users who have admin privileges for this pipeline')
-        .default([])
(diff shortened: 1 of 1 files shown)


yakanechi (approved):

minghay:
Would it be better to use a `VARCHAR` type for the adminUserIds column instead of a `TEXT` type for database column definition?  Doing that would require changing the ORM mapping to a STRING.

minghay (approved):

yk634 (approved):

